PR #190, opened 2025-10-30
chore: updated prettier configuration and vscode settings

artembelik:


.prettierignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 dist
+pnpm-lock.yaml
```
KamilEmeleev: coverage
.next

artembelik: this is already ignored by .gitignore

https://prettier.io/docs/cli#--ignore-path

PR #270, opened 2026-02-13
fix(Navbar): enhance item interaction with hover and press states

sdv43:


packages/components/src/components/Navbar/components/NavbarItem.tsx
```diff
@@ -63,7 +81,10 @@ export const NavbarItem = polymorphicForwardRef<'a', NavbarItemProps>(
               s.item,
               className
             )}
-            {...mergeProps(props, other)}
+            data-hovered={isHovered || undefined}
+            data-pressed={isActive || isPressed || undefined}
```
KamilEmeleev:
```
data-pressed={isPressed || undefined}
data-active={isActive || undefined}
```

PR #274, opened 2026-02-20
chore(ci): add Firebase PR preview workflow

KamilEmeleev:


.github/workflows/deploy-pr-preview.yml
```diff
@@ -0,0 +1,51 @@
...
+          run_install: false
+
+      - name: Install Dependencies
+        run: pnpm install
+
+      - name: Build Storybook
+        run: pnpm run build-storybook
+
+      - name: Deploy Preview
+        if: github.actor != 'dependabot[bot]'
```
artembelik: it seems better to move this check to a higher level, as there is no point in compiling the project without deployment

.github/workflows/deploy-pr-preview.yml
```diff
@@ -0,0 +1,51 @@
...
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+
+      - name: Install Node
+        uses: actions/setup-node@v4
```
artembelik:
```suggestion
        uses: actions/setup-node@v6
```

.github/workflows/deploy-pr-preview.yml
```diff
@@ -0,0 +1,51 @@
...
+  pull-requests: write
+
+jobs:
+  deploy-pr-preview:
+    runs-on: ubuntu-latest
+    if: ${{ github.event.pull_request.head.repo.full_name == github.repository }}
+
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v4
```
artembelik:
```suggestion
        uses: actions/checkout@v6
```

PR #435, opened 2026-07-30
chore: updated AGENTS.md

NikGurev:
Updated AGENTS.md according to latest review from angular developers.

It initiates common rules for stories creation, component architecture, etc.

## Summary by CodeRabbit

* **Documentation**
  * Added contributor guidance for creating Storybook stories, using accessible component props, designing compound component APIs, exporting public prop types, and following API validation and approval workflows.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -121,12 +121,10 @@ Some complex components may also contain `components/`, `utils.ts`, `intl.ts` or
 
 ### Storybook Stories
 
-- Keep each story self-contained: build any data or helper logic it needs inline inside that story's own `render`, not in shared module-level constants. The docs "Source" panel only extracts a single `export const StoryName` block, so anything defined outside it never appears there.
-- Don't pass a prop that has no effect in that story's context (e.g. `userInfo` alongside `children`, since `children` short-circuits it) and don't pass a value equal to the prop's documented `@default` (e.g. `isCompact={false}`) — both are just noise in Source.
-- Don't hand-write `argTypes` for props that already have an `as const` union type — Storybook infers the `select` control and its options automatically.
-- If a story's `render` needs hooks (`useState`, …), write it as a named function expression, `render: function Render(args) { ... }`, instead of an arrow function, so `react-hooks/rules-of-hooks` recognizes it as a component without an eslint-disable comment.
-- For compound components, add a `subcomponents` field to `meta` (see [Compound Components](#compound-components)) so the Props table documents every slot.
-- Avoid `import * as NS` for large third-party packages in stories (icon sets, etc.) — use named imports so the Storybook bundle and startup stay fast.
+- Define story data and helpers inside `render` so they appear in the Source panel.
+- Don't add `argTypes` for props inferred from component types.
+- If `render` uses hooks, use a named function: `render: function Render(args) { ... }`.
+- Add every slot of a compound component to `meta.subcomponents`.
 
 ### Styling Approach
 
@@ -139,6 +137,7 @@ Some complex components may also contain `components/`, `utils.ts`, `intl.ts` or
 
 ### Prop System
 
+- Prefer standard ARIA attributes and existing prop names to keep component APIs consistent and familiar.
 - Props with a fixed set of allowed values are exported as `as const` arrays plus a derived union type:
 
   ```ts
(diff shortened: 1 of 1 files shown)


NikGurev:
@coderabbitai review

AGENTS.md
```diff
@@ -119,6 +119,15 @@ packages/components/src/components/Button/
 
 Some complex components may also contain `components/`, `utils.ts`, `intl.ts` or `intl.json`, and `__tests__/`. Follow nearby component patterns before adding new structure.
 
+### Storybook Stories
```
KamilEmeleev, line 122: на мой взгляд, можно сократить и выкинуть некоторые пункты:

- Define story data and helpers inside `render` so they appear in the Source panel.
- Don't add `argTypes` for props inferred from component types.
- If `render` uses hooks, use a named function: `render: function Render(args) { ... }`.
- Add every slot of a compound component to `meta.subcomponents`.

AGENTS.md
```diff
@@ -154,6 +199,8 @@ Some complex components may also contain `components/`, `utils.ts`, `intl.ts` or
 - Every component `.tsx` starts with `'use client'` (Next.js RSC).
 - Keep files focused. When the main file grows, split helpers, hooks, sub-components, and translations into separate files (`utils.ts`, `intl.ts`, nested `components/`) as already done in complex components.
 - Public exports go through the component's local `index.ts`, then `packages/components/src/components/index.ts`.
+- When a render has many conditional branches, compute the values to render first (e.g. `primary`, `secondary`, `hint`), then render them once in unified markup — keep branch/selection logic separate from JSX.
```
KamilEmeleev: Предлагаю убрать эти два пункта: это общие и слишком категоричные рекомендации, а не правила проекта.

AGENTS.md
```diff
@@ -138,6 +147,9 @@ Some complex components may also contain `components/`, `utils.ts`, `intl.ts` or
   ```
 
 - All boolean props default to `false`. If `true` would be the natural default, invert the name (e.g. `hideArrow` instead of `showArrow={true}`).
+- Prefer passing through real ARIA attributes (`role`, `aria-label`, `aria-labelledby`, `aria-hidden`) over inventing custom semantic props like `label`/`decorative` — it matches patterns consumers already know, and avoids clashing with props that mean something else elsewhere in the library (e.g. `label` usually means visible text).
```
KamilEmeleev: Предлагаю оставить только первый пункт, перенести его в `Prop System` и сократить до: “Prefer standard ARIA attributes and existing prop names to keep component APIs consistent and familiar.” Второй и третий — удалить.

AGENTS.md
```diff
@@ -146,6 +158,39 @@ Some complex components may also contain `components/`, `utils.ts`, `intl.ts` or
   }
   ```
 
+### Compound Components
```
KamilEmeleev, line 157: ### Compound Components

Group related components under one public API. Consumers import the root component and access its slots as properties, for example `Component.Slot`.

```tsx
export type ComponentProps = ComponentPropsWithRef<'div'>;
export type ComponentSlotProps = ComponentPropsWithRef<'span'>;

const ComponentRoot = forwardRef<HTMLDivElement, ComponentProps>(
  (props, ref) => <div ref={ref} {...props} />
);

const ComponentSlot = forwardRef<HTMLSpanElement, ComponentSlotProps>(
  (props, ref) => <span ref={ref} {...props} />
);

type CompoundedComponent = typeof ComponentRoot & {
  Slot: typeof ComponentSlot;
};

export const Component = ComponentRoot as CompoundedComponent;
Component.Slot = ComponentSlot;
```

Export the root component and all prop types publicly. Expose slot components only through the root component.

AGENTS.md
NikGurev: Исправил

AGENTS.md
NikGurev: Изменил

AGENTS.md
NikGurev: Убрал

AGENTS.md
NikGurev: Сократил

KamilEmeleev (approved):

PR #472, opened 2026-09-07
feat(TreeSelect): add renderTag prop (DS-5277)

KamilEmeleev:
## Summary by CodeRabbit

- **New Features**
  - Added support for customizing selected tags in `TreeSelect` with the `renderTag` option.
  - Exposed `TreeSelect.Tag` for creating styled custom tags, including icons and warning variants.
  - Custom tags work across responsive and multiline layouts and respect disabled or read-only states.

- **Documentation**
  - Added usage guidance and an example demonstrating custom tag rendering.

### DIFF
--- a/packages/components/src/components/TreeSelect/TreeSelect.mdx
+++ b/packages/components/src/components/TreeSelect/TreeSelect.mdx
@@ -117,7 +117,7 @@ or `selectedTagsOverflow="multiline"` to wrap tags onto multiple lines.
 
 ### Custom tag render
 
-Use `renderTag` to customize selected tags.
+Use `renderTag` to customize tags in multiple selection mode.
 
 <Story of={Stories.CustomTagRender} />
 
--- a/packages/components/src/components/TreeSelect/TreeSelect.mdx
+++ b/packages/components/src/components/TreeSelect/TreeSelect.mdx
@@ -115,6 +115,12 @@ or `selectedTagsOverflow="multiline"` to wrap tags onto multiple lines.
 
 <Story of={Stories.SelectedTagsOverflow} />
 
+### Custom tag render
+
+Use `renderTag` to customize selected tags.
+
+<Story of={Stories.CustomTagRender} />
+
 ### Disabled
 
 When TreeSelect is disabled, it cannot be interacted with.
--- a/packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
+++ b/packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
@@ -26,7 +26,7 @@ const meta = {
     'TreeSelect.Tag': TreeSelect.Tag,
   },
   argTypes: {},
-  tags: ['status:new', 'date:2026-06-26'],
+  tags: ['status:updated', 'date:2026-09-07'],
 } satisfies Meta<typeof TreeSelect>;
 
 export default meta;
--- a/packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
+++ b/packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
@@ -1,7 +1,7 @@
 import { useState } from 'react';
 
 import { type Key, useBoolean } from '@koobiq/react-core';
-import { IconCrosshairs16 } from '@koobiq/react-icons';
+import { IconCrosshairs16, IconFolder16 } from '@koobiq/react-icons';
 import { Collection } from '@koobiq/react-primitives';
 import type { Meta, StoryObj } from '@storybook/react';
 
@@ -23,6 +23,7 @@ const meta = {
     'TreeSelect.Item': TreeSelect.Item,
     'TreeSelect.ItemContent': TreeSelect.ItemContent,
     'TreeSelect.LoadMoreItem': TreeSelect.LoadMoreItem,
+    'TreeSelect.Tag': TreeSelect.Tag,
   },
   argTypes: {},
   tags: ['status:new', 'date:2026-06-26'],
@@ -447,6 +448,55 @@ export const SelectedTagsOverflow: Story = {
           }}
         </TreeSelect>
       </FlexBox>
+    );
+  },
+};
+
+export const CustomTagRender: Story = {
+  render: function Render() {
+    type FileNode = {
+      id: number;
+      title: string;
+      children: FileNode[];
+    };
+
+    const items: FileNode[] = [
+      {
+        id: 1,
+        title: 'app',
+        children: [{ id: 2, title: 'Http', children: [] }],
+      },
+      { id: 3, title: 'config', children: [] },
+      { id: 4, title: 'public', children: [] },
+    ];
+
+    return (
+      <TreeSelect
+        items={items}
+        label="Project folders"
+        selectionMode="multiple"
+        style={{ inlineSize: 320 }}
+        placeholder="Select folders"
+        defaultValue={[2, 3]}
+        renderTag={(item, tagProps) => (
+          <TreeSelect.Tag
+            {...tagProps}
+            variant="warning-fade"
+            icon={<IconFolder16 />}
+          >
+            {item.textValue}
+          </TreeSelect.Tag>
+        )}
+      >
+        {function renderItem(item) {
+          return (
+            <TreeSelect.Item id={item.id} textValue={item.title}>
+              <TreeSelect.ItemContent>{item.title}</TreeSelect.ItemContent>
+              <Collection items={item.children}>{renderItem}</Collection>
+            </TreeSelect.Item>
+          );
+        }}
+      </TreeSelect>
     );
   },
 };
--- a/packages/components/src/components/TreeSelect/TreeSelect.test.tsx
+++ b/packages/components/src/components/TreeSelect/TreeSelect.test.tsx
@@ -369,6 +369,24 @@ describe('TreeSelect', () => {
   });
 
   describe('renderTag', () => {
+    const getTag = (key: number) => screen.getByTestId(`tag-${key}`);
+    const queryTag = (key: number) => screen.queryByTestId(`tag-${key}`);
+
+    const getRemoveButton = (key: number) =>
+      within(getTag(key)).getByRole('button', { hidden: true });
+
+    const renderTag: NonNullable<
+      TreeSelectProps<FileNode, 'multiple'>['renderTag']
+    > = (item, tagProps) => (
+      <TreeSelect.Tag
+        {...tagProps}
+        icon={<span data-testid={`tag-icon-${item.key}`} />}
+        data-testid={`tag-${item.key}`}
+      >
+        {item.textValue}
+      </TreeSelect.Tag>
+    );
+
     it('should ignore renderTag in single selection mode', () => {
       const renderTag = vi.fn(() => <div>Custom tag</div>);
 
@@ -381,6 +399,44 @@ describe('TreeSelect', () => {
     describe.each(['responsive', 'multiline'] as const)(
       '%s overflow',
       (selectedTagsOverflow) => {
+        it('should support custom elements and forward props for overflow measurement', () => {
+          const renderCustomTag = vi.fn<
+            NonNullable<TreeSelectProps<FileNode, 'multiple'>['renderTag']>
+          >((item, { className, ref, 'aria-hidden': ariaHidden }) => (
+            <div
+              ref={ref}
+              className={className}
+              aria-hidden={ariaHidden}
+              data-testid={`tag-${item.key}`}
+            >
+              Custom: {item.textValue}
+            </div>
+          ));
+
+          renderTreeSelect({
+            selectionMode: 'multiple',
+            defaultValue: [2, 7],
+            selectedTagsOverflow,
+            renderTag: renderCustomTag,
+          });
+
+          expect(getTag(2)).toHaveTextContent('Custom: Http');
+          expect(getTag(7)).toHaveTextContent('Custom: README.md');
+
+          const tagProps = renderCustomTag.mock.calls
+            .filter(([item]) => item.key === 2)
+            .at(-1)?.[1];
+
+          expect(tagProps?.className).toEqual(expect.any(String));
+          expect(getTag(2)).toHaveClass(tagProps!.className!);
+
+          if (selectedTagsOverflow === 'responsive') {
+            expect(tagProps?.ref).toEqual(
+              expect.objectContaining({ current: getTag(2) })
+            );
+          }
+        });
+
         it('should customize tags using selected items, including collapsed descendants', () => {
           renderTreeSelect({
             selectionMode: 'multiple',
@@ -397,18 +453,27 @@ describe('TreeSelect', () => {
             ),
           });
 
-          expect(screen.getByTestId('tag-2')).toHaveTextContent('File: Http');
+          expect(getTag(2)).toHaveTextContent('File: Http');
 
-          expect(screen.getByTestId('tag-7')).toHaveTextContent(
-            'File: README.md'
-          );
+          expect(getTag(7)).toHaveTextContent('File: README.md');
 
-          expect(screen.getByTestId('tag-2')).toHaveAttribute(
-            'data-variant',
-            'warning-fade'
-          );
+          expect(getTag(2)).toHaveAttribute('data-variant', 'warning-fade');
 
-          expect(screen.queryByTestId('tag-1')).not.toBeInTheDocument();
+          expect(queryTag(1)).not.toBeInTheDocument();
+        });
+
+        it('should preserve invalid styling when customizing tag content', () => {
+          renderTreeSelect({
+            selectionMode: 'multiple',
+            defaultValue: [7],
+            selectedTagsOverflow,
+            isInvalid: true,
+            renderTag,
+          });
+
+          expect(screen.getByTestId('tag-icon-7')).toBeInTheDocument();
+          expect(getTag(7)).toHaveAttribute('data-variant', 'error-fade');
+          expect(getRemoveButton(7)).toHaveAttribute('data-variant', 'error');
         });
 
         it('should remove a custom tag without opening the dropdown', async () => {
@@ -421,56 +486,50 @@ describe('TreeSelect', () => {
             selectedTagsOverflow,
             onChange,
             onOpenChange,
-            renderTag: (item, tagProps) => (
-              <TreeSelect.Tag {...tagProps} data-testid={`tag-${item.key}`}>
-                {item.textValue}
-              </TreeSelect.Tag>
-            ),
+            renderTag,
           });
 
-          await userEvent.click(
-            within(screen.getByTestId('tag-2')).getByRole('button', {
-              hidden: true,
-            })
-          );
+          await userEvent.click(getRemoveButton(2));
 
           expect(onChange).toHaveBeenCalledExactlyOnceWith([7]);
-          expect(screen.queryByTestId('tag-2')).not.toBeInTheDocument();
-          expect(screen.getByTestId('tag-7')).toHaveTextContent('README.md');
+          expect(queryTag(2)).not.toBeInTheDocument();
+          expect(getTag(7)).toHaveTextContent('README.md');
           expect(onOpenChange).not.toHaveBeenCalled();
           expect(queryPopover()).not.toBeInTheDocument();
         });
 
-        it.each(['isDisabled', 'isReadOnly'] as const)(
-          'should prevent custom tag removal when %s',
-          async (state) => {
+        it.each([
+          { isDisabled: true, isReadOnly: false },
+          { isDisabled: false, isReadOnly: true },
+        ] as const)(
+          'should prevent custom tag removal with %j',
+          async (states) => {
             const onChange = vi.fn();
 
             renderTreeSelect({
               selectionMode: 'multiple',
               defaultValue: [7],
               selectedTagsOverflow,
-              [state]: true,
+              ...states,
               onChange,
-              renderTag: (item, tagProps) => (
-                <TreeSelect.Tag {...tagProps} data-testid={`tag-${item.key}`}>
-                  {item.textValue}
-                </TreeSelect.Tag>
-              ),
+              renderTag,
             });
 
-            const tag = screen.getByTestId('tag-7');
-
-            const removeButton = within(tag).getByRole('button', {
-              hidden: true,
-            });
+            const tag = getTag(7);
+            const removeButton = getRemoveButton(7);
 
             expect(removeButton).toHaveAttribute('aria-disabled', 'true');
+
+            if (states.isDisabled) {
+              expect(tag).toHaveAttribute('data-disabled', 'true');
+            } else {
+              expect(tag).not.toHaveAttribute('data-disabled');
+            }
 
             await userEvent.click(removeButton);
 
             expect(onChange).not.toHaveBeenCalled();
-            expect(tag).toHaveTextContent('README.md');
+            expect(tag).toBeInTheDocument();
           }
         );
       }
--- a/packages/components/src/components/TreeSelect/TreeSelect.test.tsx
+++ b/packages/components/src/components/TreeSelect/TreeSelect.test.tsx
@@ -368,6 +368,115 @@ describe('TreeSelect', () => {
     });
   });
 
+  describe('renderTag', () => {
+    it('should ignore renderTag in single selection mode', () => {
+      const renderTag = vi.fn(() => <div>Custom tag</div>);
+
+      renderTreeSelect({ defaultValue: 7, renderTag });
+
+      expect(getControl()).toHaveTextContent('README.md');
+      expect(renderTag).not.toHaveBeenCalled();
+    });
+
+    describe.each(['responsive', 'multiline'] as const)(
+      '%s overflow',
+      (selectedTagsOverflow) => {
+        it('should customize tags using selected items, including collapsed descendants', () => {
+          renderTreeSelect({
+            selectionMode: 'multiple',
+            defaultValue: [2, 7],
+            selectedTagsOverflow,
+            renderTag: (item, tagProps) => (
+              <TreeSelect.Tag
+                {...tagProps}
+                variant="warning-fade"
+                data-testid={`tag-${item.key}`}
+              >
+                File: {item.value?.title}
+              </TreeSelect.Tag>
+            ),
+          });
+
+          expect(screen.getByTestId('tag-2')).toHaveTextContent('File: Http');
+
+          expect(screen.getByTestId('tag-7')).toHaveTextContent(
+            'File: README.md'
+          );
+
+          expect(screen.getByTestId('tag-2')).toHaveAttribute(
+            'data-variant',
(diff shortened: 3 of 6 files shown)


packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
```diff
@@ -451,6 +452,55 @@ export const SelectedTagsOverflow: Story = {
...
+      <TreeSelect
+        items={items}
+        label="Project folders"
+        selectionMode="multiple"
+        style={{ inlineSize: 320 }}
+        placeholder="Select folders"
+        defaultValue={[2, 3]}
+        renderTag={(item, tagProps) => (
+          <TreeSelect.Tag
+            {...tagProps}
```
lskramarov, line 483: **A consumer `slotProps` silently makes the tag unremovable and plants a focusable control inside an `aria-hidden` subtree.**

`tagProps.slotProps` carries the entire removal wiring — `{ removeIcon: { as: 'div', tabIndex: undefined, onPress: onRemove, isDisabled: isReadOnly || isDisabled } }` (`SelectedTagsResponsive.tsx:60-67`). Because JSX later-prop-wins and `slotProps` is a nested object, a completely natural override *shallow-replaces* it:

```tsx
<TreeSelect.Tag {...tagProps} slotProps={{ icon: { className: 'mine' } }}>
```

I ran this. Result:

- the remove control falls back to `IconButton`'s default `as='button'` → `tagName: BUTTON`, `tabindex="0"` — a **keyboard-focusable element inside the `aria-hidden="true"` container** (`SelectedTagsResponsive.tsx:39`), which is an axe `aria-hidden-focus` violation;
- `onPress` is gone, so clicking the × does nothing — `onChange` was never called;
- the `isReadOnly` guard is gone too.

All of it type-clean and silent. The same hazard applies to `className` (dropping `s.tag` disables the `[aria-hidden='true'] { position: absolute; visibility: hidden }` rule in `SelectedTags.module.css:40-46`).

Worth either documenting "spread `tagProps` and don't replace `slotProps`/`className`", or — better — having `SelectedTags` provide the wiring via context so `TreeSelect.Tag` merges instead of being clobbered.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
```diff
@@ -368,6 +368,115 @@ describe('TreeSelect', () => {
...
+            const removeButton = within(tag).getByRole('button', {
+              hidden: true,
+            });
+
+            expect(removeButton).toHaveAttribute('aria-disabled', 'true');
+
+            await userEvent.click(removeButton);
+
+            expect(onChange).not.toHaveBeenCalled();
+            expect(tag).toHaveTextContent('README.md');
```
lskramarov: **This assertion cannot fail.**

`tag` is captured at line 462, *before* the click at line 470. jest-dom's `toHaveTextContent` reads `node.textContent` and has no `isConnected`/`document.contains` check, and React leaves a detached subtree's own children intact — so `tag.textContent` is still `'README.md'` whether or not the tag was removed.

Verified: running the *enabled* path (so removal really happens), `expect(tag).not.toBeInTheDocument()` **and** `expect(tag).toHaveTextContent('README.md')` both pass.

So the "the tag survived" half of the test is a no-op and the case rests entirely on line 472. The sibling test at line 438 does it correctly.

```suggestion
            expect(tag).toBeInTheDocument();
```

packages/components/src/components/TreeSelect/TreeSelect.mdx
```diff
@@ -115,6 +115,12 @@ or `selectedTagsOverflow="multiline"` to wrap tags onto multiple lines.
 
 <Story of={Stories.SelectedTagsOverflow} />
 
+### Custom tag render
+
+Use `renderTag` to customize selected tags.
```
lskramarov: **Worth stating that `tagProps` must be spread onto the returned element — dropping it breaks the default overflow mode silently.**

The signature is `(item, tagProps) => ReactNode`, which reads as "`tagProps` is optional sugar". It isn't: under the default `selectedTagsOverflow="responsive"` the collapse machinery lives entirely inside it.

- `ref: itemsRefs[i]` (`SelectedTagsResponsive.tsx:54`) feeds `useHideOverflowItems`. Lose it and `itemsSizes[i]` stays `0`, so `hideByIndex` early-returns for every index — nothing ever collapses, the "+N" summary never appears, and tags are clipped by `.base { overflow: hidden }`.
- `className: s.tag` owns both the `margin-inline-end` that measurement reads and the `:is(.tag,.more)[aria-hidden='true'] { position: absolute; visibility: hidden }` rule. Override it and tags marked hidden stay visible *alongside* the "+N" counter.

`renderTag={(item) => <MyChip>{item.textValue}</MyChip>}` compiles fine and degrades with no warning. A sentence here (and the `selectionMode="multiple"` precondition — the section right above states it for `selectedTagsOverflow`) would cover it.

packages/components/src/components/TreeSelect/types.ts
```diff
@@ -68,6 +70,8 @@ export type TreeSelectProps<
 > = {
   /** Defines how selected tags are displayed when they exceed the available space. */
   selectedTagsOverflow?: TreeSelectPropSelectedTagsOverflow;
+  /** Custom renderer for selected tags in multiple selection mode. */
+  renderTag?: (item: Node<T>, tagProps: TagProps) => ReactNode;
```
lskramarov: **`TagProps` is not reachable from the package entry, and the name is already taken by an unrelated public type.**

`packages/components/src/components/index.ts:39-42` exports `./TagGroup`, `./TagList`, `./TagInput`, `./TagAutocomplete` — but not `./Tag`. Meanwhile `TagProps` *is* already a public root export, a different one: `TagGroup/Tag.tsx:21` exports `TagProps<T> = AriaItemProps<T> & {...}`, re-exported by `TagGroup/index.ts:2` and locked in `TagGroup.api.md:57`.

So a consumer typing a standalone handler against the built `dist/index.d.ts`:

```ts
import type { TagProps } from '@koobiq/react-components';
```

resolves to the wrong, generic type. Compiled it — `error TS2314: Generic type 'TagProps' requires 1 type argument(s).` Nothing points at the reachable name. `TreeSelectTagProps` does compile clean against the same entry, so it is the type consumers actually need:

```suggestion
  renderTag?: (item: Node<T>, tagProps: TreeSelectTagProps) => ReactNode;
```

Two things I checked so this is not oversold. Declaration emit *does* preserve the alias, so the rendered public signature would name the reachable type. But it does **not** silence the new `(ae-forgotten-export) The symbol "TagProps"` line in the report — API Extractor follows the alias to its right-hand side and warns there instead; `TagInput` and `TagAutocomplete` already carry that warning from the alias alone. And that warning's "entry point index.d.ts" means TreeSelect's own local index, not the package root, so it is not itself evidence of the root-export gap — the TS2314 above is.

packages/components/src/components/TreeSelect/TreeSelect.tsx
```diff
@@ -405,6 +407,7 @@ export function TreeSelectInner<
           isRequired,
         }}
         selectedTagsOverflow={selectedTagsOverflow}
+        renderTag={renderTag}
```
lskramarov, line 410: **`renderTag` type-checks in single selection mode but is silently dropped here.**

`TreeSelectProps<T, M>` accepts `renderTag` for every `M`, yet only the `'multiple'` branch forwards it — the single branch renders `state.selectedItems[0]?.textValue`. There is no dev warning (`grep NODE_ENV` in this directory returns nothing), and the MDX section does not mention the precondition even though the section directly above it states it for the sibling `selectedTagsOverflow`. The new test at line 372 codifies the silent drop as intended.

The file already has the machinery to make this a compile error: `useTreeSelectState.ts:27-38` defines `TreeSelectValueType<M>` / `TreeSelectChangeValueType<M>`, which `TreeSelectProps` inherits — that is why `defaultValue={[2, 3]}` in single mode already errors today.

I tried the one-line constraint and type-checked the whole package:

```ts
renderTag?: M extends 'multiple'
  ? (item: Node<T>, tagProps: TagProps) => ReactNode
  : never;
```

- single mode + `renderTag` → errors, as intended
- multiple mode + `renderTag` → no false positive
- dynamic `selectionMode={someUnion}` → no error (the naked type parameter distributes, so it degrades permissively)
- across every story and test in `packages/components`, exactly **one** error surfaces: `TreeSelect.test.tsx:375`, i.e. this PR's own single-mode test

Two caveats if you take it: the message reads `not assignable to type 'undefined'`, which is correct but unhelpful; and `SelectNext/types.ts:79` carries the byte-identical loose signature, so it is worth doing both at once — or leaving the types as they are and adding a `process.env.NODE_ENV !== 'production'` guarded `logger.warn` instead.

packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
```diff
@@ -451,6 +452,55 @@ export const SelectedTagsOverflow: Story = {
...
+            variant="warning-fade"
```
lskramarov, line 484: **The documented example teaches an override that permanently suppresses the invalid state.**

`SelectedTags` derives the tag colour from form state — `variant: isInvalid ? 'error-fade' : 'contrast-fade'` (`SelectedTagsResponsive.tsx:59`, `SelectedTagsMultiline.tsx:41`). Placing a literal `variant` *after* `{...tagProps}` pins it, so a field that fails validation keeps warning-coloured tags while its border and error message go red. It carries into the remove button too, via `matchTagVariantToIconButton[variant]` (`Tag.tsx:64`).

The test at lines 406-409 asserts the override wins, and `isInvalid` appears exactly once in the whole test file (line 283) — outside the `renderTag` block — so the interaction is untested.

This is copied from `SelectNext/Select.stories.tsx:459-463`, so it is propagated rather than introduced; that makes it a two-component docs problem worth fixing once. Either note the trade-off in the MDX, or have `SelectedTags` apply state-derived visuals *after* the consumer's props.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
```diff
@@ -368,6 +368,115 @@ describe('TreeSelect', () => {
...
+    it('should ignore renderTag in single selection mode', () => {
+      const renderTag = vi.fn(() => <div>Custom tag</div>);
+
+      renderTreeSelect({ defaultValue: 7, renderTag });
+
+      expect(getControl()).toHaveTextContent('README.md');
+      expect(renderTag).not.toHaveBeenCalled();
+    });
+
+    describe.each(['responsive', 'multiline'] as const)(
```
lskramarov, line 399: **No test pins the `ReactNode` half of the contract.**

Every callback in this block returns `<TreeSelect.Tag {...tagProps}>`, and `Tag` absorbs `ref` / `className` / `aria-hidden` into its root div — so the assertions would pass identically if `SelectedTags` stopped supplying them.

The reference implementation covers exactly this: `SelectNext/__tests__/SelectMultiple.test.tsx:173-192` returns a plain `<div>` built from a partial destructure of `{ className, ref, 'aria-hidden': ariaHidden }`.

I injected the regression (making `SelectedTags*` render only nodes whose type is `Tag`):

```
TreeSelect.test.tsx      ->  66 passed (66)        # green
SelectMultiple.test.tsx  ->  1 failed | 9 passed
```

CI as a whole still catches it, because both components share `SelectedTags` — so this is a suite-symmetry gap rather than a repo-level hole. Mirroring SelectNext's test here is cheap, and it would also pin the `ref` / `className` contract that the responsive mode depends on.

(Separately: the `describe.each` over both overflow modes *is* earning its keep. I mutated `SelectedTagsMultiline` alone and only the multiline arm failed, so the two arms are not redundant despite rendering identical markup in jsdom.)

packages/components/src/components/TreeSelect/TreeSelect.tsx
```diff
@@ -496,3 +500,4 @@ export const TreeSelect = TreeSelectComponent as CompoundedComponent;
 TreeSelect.Item = Tree.Item;
 TreeSelect.ItemContent = Tree.ItemContent;
 TreeSelect.LoadMoreItem = Tree.LoadMoreItem;
+TreeSelect.Tag = Tag;
```
lskramarov, line 503: **`TreeSelect.Tag` is the first slot here whose component type no api report covers.**

The other three slots are safe: `Tree` is in `tools/api-extractor/config.json`, so `Tree.api.md` backs `Item` / `ItemContent` / `LoadMoreItem`. `Tag` is guarded nowhere — it is absent from that `components` array, and the compound itself is emitted opaquely (`export const TreeSelect: CompoundedComponent_2;`, with `CompoundedComponent_2` flagged as a forgotten export), so no slot appears in TreeSelect's report either.

I mutated the built `Tag` types to check what that costs — renamed `variant`, deleted `icon` and `allowsRemoving`, and cut `tagPropVariant` from four values to one. Re-running API Extractor for TreeSelect produced a report **identical to the committed one**, so `pnpm check-api` stays green while every `renderTag` consumer — and this PR's own story, which passes `variant` and `icon` — breaks at compile time.

Adding `"Tag"` to the `components` array closes it, and covers `SelectNext.Tag` at the same time.

Worth knowing while you are here: `SelectNext` is missing from that array too, so `tools/public_api_guard/components/SelectNext.api.md` is an orphaned report that is never regenerated or checked — it still has neither `renderTag` nor `SelectNextTagProps`.

packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
```diff
@@ -451,6 +452,55 @@ export const SelectedTagsOverflow: Story = {
...
 
+export const CustomTagRender: Story = {
+  render: function Render() {
+    type FileNode = {
+      id: number;
+      title: string;
+      children: FileNode[];
+    };
+
+    const items: FileNode[] = [
```
lskramarov, line 463: **This fixture shadows the module-level `items` and reuses its ids for different nodes.**

`TreeSelect.stories.tsx:35` already defines `items`, which the other dynamic stories in this file share. The local one shadows it, renames the child key `test` to `children`, and reassigns ids: `3` is `index.html` (nested under `Http`) in the module fixture but `config` (a root) here, and `4` is `Providers` vs `public`. Since the MDX renders story source verbatim, a reader comparing "Selected value tags" with "Custom tag render" has to re-derive which tree is which.

The module fixture already contains `app`(1) → `Http`(2), `config`(6) and `public`(9), so `items={items}` with `defaultValue={[2, 6]}` reproduces this story — nested selection included — and drops ~15 lines along with the `FileNode` type and the `children: []` boilerplate it forces.

packages/components/src/components/TreeSelect/types.ts
```diff
@@ -43,6 +44,7 @@ import type { TreeCollection } from './TreeInner';
 export type TreeSelectItemProps = TreeItemProps;
 export type TreeSelectItemContentProps = TreeItemContentProps;
 export type TreeSelectLoadMoreItemProps = TreeLoadMoreItemProps;
+export type TreeSelectTagProps = TagProps;
```
lskramarov, line 47: **Third verbatim copy of this signature.**

The identical declaration — and the identical JSDoc line — now lives in three places:

- `SelectedTags/types.ts:30` (the producer; the only place it is actually called)
- `SelectNext/types.ts:79`
- here

TreeSelect only forwards the value, so nothing type-checks the copies against the producer; structural compatibility hides drift instead of failing the build. That drift has already started on the sibling prop: `SelectNext` documents `selectedTagsOverflow` with its value list and `@default 'responsive'`, while TreeSelect's copy is a bare one-liner — and per AGENTS.md the Storybook Props table is generated from those comments, so TreeSelect's docs page omits the default it actually applies.

This file already establishes the fix eleven lines below: `treeSelectPropSelectedTagsOverflow = selectedTagsPropOverflow` re-exposes a SelectedTags contract under a TreeSelect name. Exporting the prop type from `SelectedTags` (`SelectedTagsProps<T>['renderTag']`, or a named `SelectedTagsRenderTag<T>`) and referencing it here would finish the pattern — and make the new `Node` import unnecessary.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
```diff
@@ -368,6 +368,115 @@ describe('TreeSelect', () => {
...
+          const onChange = vi.fn();
+          const onOpenChange = vi.fn();
+
+          renderTreeSelect({
+            selectionMode: 'multiple',
+            defaultValue: [2, 7],
+            selectedTagsOverflow,
+            onChange,
+            onOpenChange,
+            renderTag: (item, tagProps) => (
```
lskramarov: **This five-line callback is byte-identical to the one at line 455.**

The file already keeps its shared helpers at the top — `renderTreeSelect` at line 72, and `getRoot` / `getControl` / `getClearButton` / `queryPopover` at 78-81. One hoisted `renderTag` at the top of this `describe` would cover both call sites, and a `getTag(key)` helper would replace the eight inline `getByTestId`/`queryByTestId` tag lookups in this block.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
```diff
@@ -368,6 +368,115 @@ describe('TreeSelect', () => {
...
+        it.each(['isDisabled', 'isReadOnly'] as const)(
+          'should prevent custom tag removal when %s',
+          async (state) => {
+            const onChange = vi.fn();
+
+            renderTreeSelect({
+              selectionMode: 'multiple',
+              defaultValue: [7],
+              selectedTagsOverflow,
+              [state]: true,
```
lskramarov: **The computed key drops prop-name checking.**

`state` is typed `'isDisabled' | 'isReadOnly'`, and a computed key whose type is a union of literals widens to an index signature rather than the two named properties — so the name is no longer checked against `Partial<TreeSelectProps<FileNode, M>>`. Typo the `it.each` array (`'isReadonly'`) or rename the prop, and this still compiles, still renders, and still passes while asserting nothing about the path it claims to cover.

It is also the only `[x]: true` spread in any `*.test.tsx` in the repo — `SelectNext/__tests__/SelectMultiple.test.tsx:139-162` writes the literals out. `it.each([{ isDisabled: true }, { isReadOnly: true }] as const)` keeps the parameterization and the checking.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
```diff
@@ -368,6 +368,115 @@ describe('TreeSelect', () => {
...
+
+          expect(screen.getByTestId('tag-2')).toHaveAttribute(
+            'data-variant',
+            'warning-fade'
+          );
+
+          expect(screen.queryByTestId('tag-1')).not.toBeInTheDocument();
+        });
+
+        it('should remove a custom tag without opening the dropdown', async () => {
```
lskramarov, line 479: **Removing a tag does not work on the first click while the dropdown is open, and this is the only removal test.**

The title scopes it to the closed dropdown, which is the path that works. With the popover open the click on a tag's ✕ is consumed by the overlay's dismiss instead:

```
selectionMode="multiple", defaultValue={[2, 7]}, defaultOpen, renderTag
click ✕ inside tag-2  ->  onChange: 0 calls
                          onOpenChange: [false]
                          tag-2 still in the DOM
```

Only after the popover finishes unmounting does a second click produce `onChange([7])`. It behaves identically with default tags, so this is pre-existing rather than something `renderTag` introduced — but it is exactly the flow a reader of the new `CustomTagRender` story will hit first (open the list, change your mind, click ✕, nothing happens), and eight green tests around the closed path give the opposite impression.

Worth either a companion test asserting the open-dropdown behaviour you actually want, or a fix so the tag's remove control does not lose its press to the dismiss handler.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
```diff
@@ -368,6 +368,115 @@ describe('TreeSelect', () => {
...
+          );
+
+          expect(onChange).toHaveBeenCalledExactlyOnceWith([7]);
+          expect(screen.queryByTestId('tag-2')).not.toBeInTheDocument();
+          expect(screen.getByTestId('tag-7')).toHaveTextContent('README.md');
+          expect(onOpenChange).not.toHaveBeenCalled();
+          expect(queryPopover()).not.toBeInTheDocument();
+        });
+
+        it.each(['isDisabled', 'isReadOnly'] as const)(
```
lskramarov: **These two states render differently, but the shared assertions cannot tell them apart.**

`tagProps.isDisabled` is the field-level `isDisabled`, while only `slotProps.removeIcon.isDisabled` is `isReadOnly || isDisabled`. So:

- `isDisabled` → tag root gets `data-disabled="true"`
- `isReadOnly` → tag root has **no** `data-disabled`, just an `aria-disabled` remove button

Both branches pass on `aria-disabled` + text content alone, so a regression that starts greying read-only tags — or stops greying disabled ones — ships silently. Asserting `data-disabled` per state would pin it.

While you are in here: this block also quietly contradicts `describe('read only')` further down, whose "should hide clear and tag remove actions" does

```ts
expect(within(getControl()).queryAllByRole('button')).toHaveLength(0);
```

That passes in *every* state, because the tag container is `aria-hidden` — with `isReadOnly={false}`, `queryAllByRole('button')` returns 0 while `queryAllByRole('button', { hidden: true })` returns 2. Your new tests use the correct query and prove the remove controls are present and merely `aria-disabled`, which falsifies that test's title. Two adjacent describes now make opposite claims about the same behaviour; worth fixing the older one in passing.

(Also worth deciding deliberately: read-only currently renders a full-contrast tag with a visible ✕ affordance that does nothing.)

packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
KamilEmeleev: This PR exposes the existing `renderTag` API in TreeSelect, following Select. It does not change how `SelectedTags` builds `tagProps` or how Tag handles overrides. The example already forwards these props. Moving the wiring into context would be a separate change to the shared implementation and should be considered for both components together.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
KamilEmeleev: Agreed. Replaced this assertion with `toBeInTheDocument()` so the test verifies that the tag remains mounted after the click.

packages/components/src/components/TreeSelect/TreeSelect.mdx
KamilEmeleev: Added the multiple-selection qualifier to the description. The example already demonstrates spreading `tagProps`. The forwarding requirements come from the existing Select/SelectedTags contract; this PR preserves that contract rather than introducing a different one.

packages/components/src/components/TreeSelect/types.ts
KamilEmeleev: Agreed. The callback signature now uses the publicly exported `TreeSelectTagProps`, avoiding confusion with the unrelated `TagProps` exported by TagGroup.

packages/components/src/components/TreeSelect/TreeSelect.tsx
KamilEmeleev: Ignoring `renderTag` in single-selection mode is intentional and matches Select. The callback customizes selected tags, which are only rendered in multiple-selection mode. This is now stated in the documentation. Introducing a conditional type or a runtime warning would change the shared API convention and should be discussed separately for both components.

packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
KamilEmeleev: The explicit variant override is intentional: this story demonstrates customizing the tag’s appearance. `tagProps` provides the default state-dependent variant, and the consumer can choose to preserve or override it. Applying the error variant after consumer props would restrict that customization. The field’s validation state itself is unaffected.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
KamilEmeleev: Agreed. Added a plain `div` renderer test for both overflow modes, including checks for className forwarding and responsive measurement ref attachment.

packages/components/src/components/TreeSelect/TreeSelect.tsx
KamilEmeleev: The API coverage gap is valid, but it predates this PR: TreeSelect.Tag exposes the existing Tag component, already exposed through SelectNext.Tag, without changing its props. Adding shared Tag coverage and restoring SelectNext report generation should be handled in a separate tooling change. This PR updates TreeSelect’s own API report.

packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
KamilEmeleev: The local fixture is intentional. Our Storybook convention keeps example data inside `render` so the displayed source is self-contained. Item IDs are scoped to each collection and do not need to match IDs in other stories. This small fixture demonstrates nested selection without pulling in the larger shared tree.

packages/components/src/components/TreeSelect/types.ts
KamilEmeleev: The callback is type-checked against `SelectedTags` when it is forwarded, so incompatible signatures are already checked by TypeScript. Keeping an explicit public signature also lets it reference the consumer-facing `TreeSelectTagProps` name. Extracting a shared renderer type is a possible refactoring, but is not required for this addition. The existing `selectedTagsOverflow` documentation is separate from this change.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
KamilEmeleev: Extracted the repeated tag renderer and tag lookup helpers within the `renderTag` test block.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
KamilEmeleev: Replaced the computed property with explicit `isDisabled` / `isReadOnly` test cases. This makes the intended props clearer and keeps their names checked by TypeScript.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
KamilEmeleev: This also reproduces with default tags and without `renderTag`: the existing modal overlay consumes the first outside interaction to dismiss the dropdown. This PR does not change overlay dismissal or tag removal handling. Changing that interaction should be addressed separately; the new test is explicitly scoped to removal while the dropdown is closed.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
KamilEmeleev: Added distinct assertions to the new tests: disabled tags have `data-disabled`, while read-only tags do not; removal is disabled in both cases. The older read-only test and the existing remove affordance predate this PR, so changes to those are outside this addition.

lskramarov (approved):

PR #473, opened 2026-09-08
feat(Tag, TagGroup, TagList): reuse the shared Tag component (DS-5278)

KamilEmeleev:
## What

`TagGroup` and `TagList` now render the shared `Tag` component instead of two hand-rolled copies. Both local stylesheets and both variant maps are deleted. `Select`'s chips are repointed at the shared stylesheet because the file they used lived under `TagGroup`; nothing else in the deprecated component changes.

## Behaviour changes

- **Focus ring.** TagGroup tags no longer paint an outline on pointer focus — only on `:focus-visible`, matching TagList and the standalone `Tag`. The dead `data-focused` / `data-pressed` attributes are gone. An `error-fade` tag's keyboard outline is now `--kbq-states-line-focus-error` rather than the old variant-agnostic theme colour.
- **Tag internals are one flex context.** `gridCellProps` now lands on the shared `.body` wrapper (`display: contents`), so icon, label and remove button become direct flex children of the root. Tags with an icon change spacing and width, and `@mixin ellipsis` on the label starts working.
- **`slotProps.root` wins `style`.** TagList previously let the item's own `style` win. Both wrappers now let the slot win, matching the base `Tag`'s slot contract. Interaction-state `data-*` attributes stay authoritative and are applied after the merge.
- **`slotProps.removeIcon` wins over React Aria.** TagList previously let the hook win, so a custom `aria-label` was silently dropped. Both wrappers now put the slot last.
- **`slotProps.removeIcon.children` replaces the glyph.** It used to be ignored, because JSX children were appended after the spread.
- **TagGroup inherits the `--kbq-tag-*` override points** that TagList already had. Documented in `TagGroup.mdx`.

## Summary by CodeRabbit

- **Improvements**
  - Standardized Tag and TagList rendering for more consistent behavior across components.
  - Improved support for customizing tag roots, including custom classes, styles, refs, and focus handlers.
  - Tag removal controls now integrate more consistently with component customization.
  - Keyboard navigation between tags remains supported when custom root properties are applied.
  - Tag and TagList variants now share a consistent set of supported options.


packages/components/src/components/TagGroup/components/Tag/Tag.tsx
```diff
@@ -43,82 +32,56 @@ export function Tag<T>(props: TagProps<T>) {
...
-        s[variant],
-        isFocused && s.focused,
-        isHovered && s.hovered,
-        isDisabled && s.disabled,
-        textNormalMedium,
-        className
-      ),
-      style,
-      'data-variant': variant,
       'data-focused': isFocused || undefined,
```
lskramarov: **TagGroup tags lose their focus outline on pointer focus.**

The deleted `TagGroup/components/Tag/Tag.module.css` had:

```css
/* focused */
.focused {
  --tag-outline-color: var(--kbq-states-line-focus-theme);
}
```

applied via `isFocused && s.focused` — i.e. on **any** focus modality. The shared `Tag/Tag.module.css` only ships `:where([data-focus-visible])` rules; `grep -rn "data-focused" packages/components/src --include=*.css` returns nothing.

So `data-focused` (and `data-pressed`) are still computed and emitted here, but no rule consumes them. Click a tag with the mouse: `useSelectableItem` moves DOM focus to the row, `isFocused` is `true` and `isFocusVisible` is `false`, so the tag now paints `--tag-outline-color: transparent` and there is no visible focus affordance — even though Backspace/Delete will remove the focused tag.

Secondary: the old rule was variant-agnostic, so an `error-fade` TagGroup tag's keyboard outline silently changes from `--kbq-states-line-focus-theme` to `--kbq-states-line-focus-error`.

If dropping the pointer-focus ring is intentional (it does align TagGroup with TagList), worth calling out in the PR description and dropping the now-dead `data-focused` / `data-pressed` attributes; otherwise the shared sheet needs a `[data-focused]` rule.

packages/components/src/components/TagGroup/components/Tag/Tag.tsx
```diff
@@ -43,82 +32,56 @@ export function Tag<T>(props: TagProps<T>) {
...
-    {
-      className: s.icon,
-    },
-    slotProps?.icon
-  );
+  const removeIconProps = allowsRemoving
+    ? mergeProps<(TagRemoveButtonProps | undefined)[]>(
+        removeButtonPropsAria,
+        {
+          isDisabled,
```
lskramarov: **The remove button's `isDisabled` now overrides React Aria's own value.** The precedence here is inverted relative to `main`.

Pre-PR: `mergeProps({ isDisabled, tabIndex: -1, ... }, removeButtonPropsAria, slotProps?.removeIcon)` — aria won.
Now: `mergeProps(removeButtonPropsAria, { isDisabled, tabIndex: -1 }, slotProps?.removeIcon)` — the wrapper wins.

Those two values are computed differently. `@react-aria/tag@3.8.1` computes `removeButtonProps.isDisabled = state.disabledKeys.has(item.key) || item.props.isDisabled` with no behavior gate, whereas the `isDisabled` destructured from `useTag` comes from `useSelectableItem` (`props.isDisabled || manager.isDisabled(key)`), and `SelectionManager.isDisabled` returns `disabledBehavior === 'all' && ...`.

With `<TagGroup disabledKeys={['a']} disabledBehavior="selection" onRemove={...}>`, aria reports the remove button disabled while the wrapper reports `false`, so the remove button on a disabled tag becomes clickable.

Nothing catches it — `TagGroup.test.tsx` never passes `onRemove`, so this whole branch is dead under test.

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
```diff
@@ -83,58 +69,40 @@ export function TagItem<T extends object>(props: TagItemProps<T>) {
...
-  >(
-    {
-      isCompact: true,
-      className: s.cancelIcon,
-      variant: matchVariantToIconButton[variant],
     },
-    slotProps?.removeIcon,
-    removeButtonPropsAria
+    slotProps?.root,
+    { ref, style, className }
```
lskramarov: **Redundant `ref` here creates a brand-new callback ref on every render.**

`useTagListItem` already puts `ref` into `rowProps` (`packages/primitives/src/behaviors/useTagListItem.ts:266`), so `mergeProps` takes its `key === 'ref' && a && b` branch and calls `mergeRefs(a, b)`. `mergeRefs` only short-circuits at `refs.length === 1`, so with two populated slots it returns a fresh closure each render.

React therefore detaches (`ref.current = null`) and re-attaches the row node on **every** re-render — and every hover, focus move, press or selection toggle re-renders all N tags. A consumer passing a callback ref through `slotProps.root.ref` (e.g. `el => observer.observe(el)`) sees it fire with `null` then the node on every unrelated render, leaking observers.

TagGroup genuinely needs its trailing `ref` — `@react-aria/tag`'s `useTag` returns no ref in `rowProps` — but here it is pure redundancy.

```suggestion
    { style, className }
```

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
```diff
@@ -83,58 +69,40 @@ export function TagItem<T extends object>(props: TagItemProps<T>) {
...
-    [IconButtonProps, IconButtonProps | undefined, IconButtonProps]
-  >(
-    {
-      isCompact: true,
-      className: s.cancelIcon,
-      variant: matchVariantToIconButton[variant],
     },
-    slotProps?.removeIcon,
-    removeButtonPropsAria
+    slotProps?.root,
```
lskramarov: **`slotProps.root` moved after the computed state block, so consumers can now clobber the `data-*` attributes that drive the styling.**

Pre-PR: `mergeProps(rowProps, hoverProps, focusProps, slotProps?.root, { 'data-testid': testId, 'data-selected': isSelected || undefined, ... })` — the computed object was last and authoritative. Now the computed block is at position 4 and `slotProps?.root` at 5.

`<TagList.Tag slotProps={{ root: { 'data-selected': 'true' } }}>` now pins the tag to the `.theme-fade:where([data-selected])` block (blue background, white text, white remove-icon overrides) regardless of `state.selectionManager`, desyncing the visual selected state from the real one. Same inversion for `data-testid`, `data-hovered`, `data-pressed` and `data-focus-visible`.

If letting slot props win is deliberate, the interaction-state attributes should probably still be applied after it.

packages/components/src/components/TagGroup/components/Tag/Tag.tsx
```diff
@@ -43,82 +32,56 @@ export function Tag<T>(props: TagProps<T>) {
...
-    slotProps?.removeIcon
-  );
-
-  const contentProps = mergeProps(
-    {
-      className: s.content,
-    },
-    slotProps?.content
+    slotProps?.root,
+    { ref, style, className }
```
lskramarov: **This reverses `style` precedence in TagGroup: `slotProps.root.style` is now discarded whenever the item also sets `style`.**

`mergeProps` does not deep-merge `style` — it replaces the object wholesale (`result[key] = b !== undefined ? b : a`). The old chain was `mergeProps({ className, style, ... }, rowProps, hoverProps, focusProps, slotProps?.root)`, so `slotProps.root` was last and won.

`<Tag style={{ maxInlineSize: 100 }} slotProps={{ root: { style: { maxInlineSize: 200 } } }} />` rendered 200px before and renders 100px now, with the slot's style object discarded rather than merged.

The new test can't catch this: it sets `style` only on `slotProps.root` and leaves the item's `style` undefined, so `mergeProps` falls through to the earlier value under either order. Worth a case that sets both.

Note TagList keeps the opposite direction, so the two wrappers were inconsistent before and are now inconsistent the other way.

packages/components/src/components/Select/components/Tag/Tag.tsx
```diff
@@ -18,8 +18,8 @@ import { IconXmarkS16 } from '@koobiq/react-icons';
 
 import { utilClasses } from '../../../../styles/utility';
 import { IconButton } from '../../../IconButton';
+import s from '../../../Tag/Tag.module.css';
```
lskramarov, line 21: **Repointing at the shared stylesheet subscribes Select's chips (and TagGroup's tags) to the public `--kbq-tag-*` override variables — which also defeat the disabled state.**

The deleted `TagGroup/components/Tag/Tag.module.css` read bare private vars (`background-color: var(--tag-bg-color)`), so `.disabled` was unbeatable. `Tag/Tag.module.css` reads `background-color: var(--kbq-tag-bg-color, var(--tag-bg-color))` while `.base:where([data-disabled])` sets only the **private** `--tag-bg-color`.

Two consequences, for this file and for TagGroup:

1. An app that sets `--kbq-tag-bg-color` on a wrapper or `:root` to theme its `TagList` (documented at `TagList.mdx:126`) now also repaints every `Select` trigger chip and every `TagGroup` tag — components it never targeted.
2. Disabled Select chips and disabled TagGroup tags render in that custom colour instead of `--kbq-states-background-disabled`, because the public var wins the `var()` fallback chain. Same for `--kbq-tag-color`, `--kbq-tag-icon-color` and `--kbq-tag-outline-color`.

`TagGroup.mdx` has no CSS-variables section at all, so these newly-inherited override points are undocumented.

packages/components/src/components/TagGroup/components/Tag/Tag.tsx
```diff
@@ -43,82 +32,56 @@ export function Tag<T>(props: TagProps<T>) {
...
-      </div>
-    </div>
+    <BaseTag
+      {...rootProps}
+      icon={icon}
+      variant={variant}
+      allowsRemoving={allowsRemoving}
+      isDisabled={isDisabled}
+      slotProps={{
+        body: gridCellProps,
```
lskramarov, line 81: **Injecting `gridCellProps` through the `body` slot puts `.body { display: contents }` on the `role="gridcell"` element.**

Pre-PR both wrappers rendered an unclassed `<div {...gridCellProps}>` — a real block box. `Tag/Tag.module.css:36-38` sets `display: contents`, and `BaseTag` merges that class into `slotProps.body`, so the gridcell now generates no box.

Two consequences:

1. **Accessibility** — `display: contents` on an element carrying an explicit ARIA role is the classic "dropped from the accessibility tree" pattern. If any target engine drops it, the `grid > row > gridcell` ownership both components rely on breaks. Worth a VoiceOver/NVDA pass before merge.
2. **Layout** — icon, label and remove button become direct flex items of `.base`, so `gap: var(--kbq-size-3xs)` now applies on top of the existing `.icon` / `.content` / `.cancelIcon` margins, and `@mixin ellipsis` on `.content` starts working (it was inert on an inline box). Every TagGroup/TagList tag with an icon changes width and inner spacing.

This looks like the intended fix, but it is an unflagged, untested visual change to two shipped components — worth a before/after screenshot in the PR description. Same line at `TagList/components/TagItem/TagItem.tsx:98`.

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
```diff
@@ -1,26 +1,14 @@
 import { useRef } from 'react';
```
lskramarov, line 3: **Missing `'use client'`.**

Commit `fce8c3df` added it to the structurally identical sibling `TagGroup/components/Tag/Tag.tsx:1` in this same PR, but this file — refactored into the same shape, calling `useRef`, `useTagListItem`, `useFocusRing` and `useHover` — still starts with the import.

`AGENTS.md` → Coding Conventions: "Every component `.tsx` starts with `'use client'` (Next.js RSC)", and Component Anatomy: "`'use client'` is the first line."

It works today only because `TagList.tsx` / `TagListInner.tsx` carry the boundary above it, so a direct import from an RSC tree fails in the consumer's app rather than in this repo's CI. If the directive was needed on the TagGroup side, it's needed here too.

```suggestion
'use client';

import { useRef } from 'react';
```

packages/components/src/components/TagGroup/TagGroup.test.tsx
```diff
@@ -66,6 +67,46 @@ describe('TagGroup', () => {
   describe('check Tag', () => {
     const getTag = () => screen.getByTestId(TAG_GROUP__TEST_ID);
 
+    it('should apply root slot props and preserve keyboard navigation', async () => {
```
lskramarov, line 99: **The TagGroup remove-button path has no test coverage, even though this PR rewired every one of its props.**

`grep -c onRemove TagGroup.test.tsx` returns `0`, so `allowsRemoving` is always `false` and the entire `removeIconProps` branch (`TagGroup/components/Tag/Tag.tsx:59-68`) is dead under test. The PR changed that merge from `mergeProps({isDisabled, tabIndex:-1, isCompact, className, variant}, removeButtonPropsAria, slotProps?.removeIcon)` to `mergeProps(removeButtonPropsAria, {isDisabled, tabIndex:-1}, slotProps?.removeIcon)` and moved `isCompact` / `size` / `className` / `variant` / `aria-label` / `children` into `BaseTag`.

A broken `aria-label`, a button that never renders, one that stays enabled on a disabled tag, or one that lands in the tab order would all ship green. TagList has three such tests; TagGroup has none.

Compounding it, the only disabled-flavoured test in this file (line 122) asserts nothing — nothing in that fixture is disabled and nothing was ever focused, so the expectation is unconditionally true:

```js
it('should be disabled', () => {
  render(renderComponent({}));

  expect(getTag()).not.toHaveFocus();
});
```

Slot pass-through for `icon`, `content` and `removeIcon` is untested in both files too — only `root` is covered.

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
```diff
@@ -83,58 +69,40 @@ export function TagItem<T extends object>(props: TagItemProps<T>) {
...
-    removeButtonPropsAria
+    slotProps?.root,
+    { ref, style, className }
   );
 
-  const contentProps = mergeProps({ className: s.content }, slotProps?.content);
-  const iconProps = mergeProps({ className: s.icon }, slotProps?.icon);
+  const removeIconProps = allowsRemoving
+    ? mergeProps<[TagRemoveButtonProps | undefined, TagRemoveButtonProps]>(
+        slotProps?.removeIcon,
```
lskramarov: **The two wrappers now feed the same shared `removeIcon` slot with opposite precedence.**

Here `mergeProps(slotProps?.removeIcon, removeButtonPropsAria)` puts the hook's props last, so `useTagListItem`'s `aria-label`, `id`, `tabIndex` and `isDisabled` overwrite the consumer's. `TagGroup/components/Tag/Tag.tsx:60-67` uses `mergeProps(removeButtonPropsAria, {...}, slotProps?.removeIcon)`, so there the consumer wins.

Concretely, `<TagList.Tag slotProps={{ removeIcon: { 'aria-label': 'Custom' } }}>` keeps the default label, while the identical prop on `<TagGroup><Tag>` applies. It also contradicts the base component's own contract, which is written as `mergeProps({...defaults}, slotProps?.removeIcon)` — slot wins.

The asymmetry predates this PR, but the refactor merged both into one shared component and one shared type without unifying it, so it now reads as a single documented API that behaves differently depending on the parent. Worth picking one order for both.

packages/components/src/components/Tag/Tag.tsx
```diff
@@ -35,44 +35,41 @@ export const Tag = forwardRef<ComponentRef<'div'>, TagProps>((props, ref) => {
...
-    slotProps?.removeIcon
-  );
+  const removeButtonProps = allowsRemoving
+    ? mergeProps<[TagRemoveButtonProps, TagRemoveButtonProps | undefined]>(
+        {
+          isDisabled,
+          isCompact: true,
+          size: 'l',
+          className: s.cancelIcon,
+          children: <IconXmarkS16 />,
```
lskramarov, line 59: **Moving the glyph into `children` means `slotProps.removeIcon.children` now replaces it — it used to be ignored.**

Pre-PR both wrappers rendered `<IconButton size="l" {...removeButtonProps}><IconXmarkS16 /></IconButton>`. The JSX transform appends element children *after* the spread, so any `children` a consumer put in `slotProps.removeIcon` was discarded. Now `children` is a default in this merge and `slotProps?.removeIcon` is merged after it (line 63), and both wrappers forward the consumer's `slotProps.removeIcon` into that slot.

So `<TagList.Tag slotProps={{ removeIcon: { children: null } }}>` previously still showed the cross and now renders a remove button with no icon at all.

Plausibly desirable — but it is an unannounced public behaviour change for `TagList.Tag` and TagGroup's `Tag`, and it is uncovered.

packages/components/src/components/Select/components/Tag/Tag.tsx
```diff
@@ -58,13 +58,9 @@ export const Tag = forwardRef<ComponentRef<'div'>, TagProps>((props, ref) => {
...
   const rootProps = mergeProps({
-    className: clsx(
-      s.base,
-      s[variant],
-      isDisabled && s.disabled,
-      textNormalMedium,
-      className
-    ),
+    className: clsx(s.base, s[variant], textNormalMedium, className),
+    'data-disabled': isDisabled || undefined,
```
lskramarov, line 62: **These two `data-*` keys sit before `...other`, so a caller can override them** — the opposite of the shared `Tag` this file is being aligned with (`Tag/Tag.tsx:38` merges `other` **first**).

`rootProps` is a plain object literal here (the enclosing `mergeProps` has a single argument, so it is just a shallow clone — as are the `contentProps` and `iconProps` calls below). `className` is computed *before* the spread and keeps `s[variant]`, so a caller passing `data-variant` through the `ExtendableComponentPropsWithRef<..., 'div'>` surface can render `data-variant="theme-fade"` on an element wearing the `.contrast-fade` class, and can switch `.base:where([data-disabled])` on for an enabled tag (or off for a disabled one) while `IconButton` still receives the real `isDisabled`.

This also narrows a guarantee the PR removed: `isDisabled && s.disabled` was a computed class no caller could override. Today's two call sites only pass `className` / `ref` / `aria-hidden` / `isDisabled` / `variant` / `onRemove`, so nothing breaks yet.

packages/components/src/components/Select/components/Tag/Tag.tsx
```diff
@@ -18,8 +18,8 @@ import { IconXmarkS16 } from '@koobiq/react-icons';
 
 import { utilClasses } from '../../../../styles/utility';
 import { IconButton } from '../../../IconButton';
+import s from '../../../Tag/Tag.module.css';
 import type { TagGroupPropVariant } from '../../../TagGroup';
-import s from '../../../TagGroup/components/Tag/Tag.module.css';
 
 import intlMessages from './intl.json';
 import { matchVariantToCloseButton } from './utils';
```
lskramarov, line 25: **The PR deleted the TagGroup and TagList copies of this variant map but left Select's byte-identical fork.**

I byte-compared them: `Select/components/Tag/utils.ts`'s `matchVariantToCloseButton` is value-for-value identical to `matchTagVariantToIconButton` in `Tag/utils.ts`, and `Select/components/Tag/intl.json` is byte-identical to `Tag/intl.json`.

So the consolidation is two-thirds done. The next variant added to `tagPropVariant` will update TagGroup, TagList and SelectNext while Select's chips keep the old mapping, and a wording or locale fix to the shared `remove` label silently skips Select's close button — leaving two different accessible names for the same control. `TagGroupPropVariant` is now literally `TagPropVariant`, so the two maps are keyed by the same union and cannot legitimately differ.

Suggest deleting both files and importing `matchTagVariantToIconButton` and the intl bundle from `../../../Tag`.

Bigger question, since this file already borrows the shared stylesheet: could it just render `<Tag>` with `slotProps={{ removeIcon: { as: 'div', tabIndex: undefined, onPress: onRemove } }}` and drop the hand-rolled markup entirely? The `.body` wrapper is `display: contents`, so it is layout-neutral.

packages/components/src/components/TagList/TagList.test.tsx
```diff
@@ -62,6 +62,46 @@ describe('TagList', () => {
...
+
+    const [first, second] = screen.getAllByRole('row');
+
+    expect(ref.current).toBe(first);
+    expect(first).toHaveClass('tag', 'slot');
+    expect(first).toHaveStyle({ padding: '20px' });
+
+    await user.tab();
+
+    expect(first).toHaveFocus();
```
lskramarov: **These focus assertions omit the `waitFor` the rest of this file uses.**

With roving tabindex `focusedKey` is `null` on mount, so every row gets `tabIndex: -1` (`useTagListItem.ts:269`) and the grid container gets `0`. `user.tab()` therefore focuses the **container**; React Aria then sets `focusedKey` and a `useEffect` (`ref.current?.focus()`, `useTagListItem.ts:145-154`) moves DOM focus onto the row on a later commit. Every other focus assertion in this file wraps that in `waitFor` (14 occurrences); these don't, so they rely on user-event's `act` wrapper flushing the effect and will flake if the hand-off ever takes an extra tick.

Two more nits in the same test:

- `expect(onFocus).toHaveBeenCalledTimes(1)` measures React Aria internals rather than slot-prop application. The test omits `onRemove`, so no remove button exists; add one (the normal configuration) and `focusin` bubbling makes the count 2 with no user-visible change.
- `expect(ref.current).toBeNull()` after `unmount()` can only fail if React or `@react-aria/utils` breaks — the real regression (slot ref not merged, or clobbered by the trailing `{ ref }`) is already caught by the `toBe(first)` assertion above.

Same shape at `TagGroup.test.tsx:98`.

packages/components/src/components/TagGroup/components/Tag/Tag.tsx
```diff
@@ -43,82 +32,56 @@ export function Tag<T>(props: TagProps<T>) {
...
   );
 
-  const iconProps = mergeProps(
-    {
-      className: s.icon,
-    },
-    slotProps?.icon
-  );
+  const removeIconProps = allowsRemoving
+    ? mergeProps<(TagRemoveButtonProps | undefined)[]>(
```
lskramarov: **Array generic instead of a tuple erases per-argument type checking here.**

With `T = (TagRemoveButtonProps | undefined)[]`, TypeScript coerces every argument to the same union, so React Aria's `removeButtonProps` shape is never validated positionally against `IconButtonProps`. Both sibling call sites use the tuple form:

- `TagList/components/TagItem/TagItem.tsx:85` — `mergeProps<[TagRemoveButtonProps | undefined, TagRemoveButtonProps]>`
- `Tag/Tag.tsx:53` — `mergeProps<[TagRemoveButtonProps, TagRemoveButtonProps | undefined]>`

Compounding it, both wrappers switched from the closed `IconButtonProps` they used pre-PR to `TagRemoveButtonProps = IconButtonProps<ElementType>`, whose polymorphic `ElementType` parameter widens the accepted shape considerably. Between the two, a typo (`isCompat`, `varaint`, `tabindex`) or a prop a future React Aria bump adds that `IconButton` doesn't consume compiles silently and gets spread onto the DOM as an unknown attribute.

```suggestion
    ? mergeProps<[TagRemoveButtonProps, TagRemoveButtonProps, TagRemoveButtonProps | undefined]>(
```

packages/components/src/components/TagGroup/components/Tag/Tag.tsx
KamilEmeleev: Intentional — TagGroup is being aligned with TagList and the standalone `Tag`, neither of which ever had a pointer-focus ring. Dropped the now-dead `data-focused` / `data-pressed` attributes from both wrappers.

The `error-fade` change is intentional too: the variant-specific `--kbq-states-line-focus-error` is what the shared sheet ships, and the old variant-agnostic rule was the outlier. Both are listed in the PR description as behaviour changes.

packages/components/src/components/TagGroup/components/Tag/Tag.tsx
KamilEmeleev: Fixed — the wrapper no longer computes `isDisabled` for the remove button, so React Aria's value stands.

Worth noting it isn't reachable through TagGroup's public API today: `AriaTagGroupProps` extends `MultipleSelection`, which has no `disabledBehavior`, so the two values can't diverge from outside. That means no failing-before regression test is possible — added the reachable coverage instead, see the new `removable tags` suite in `TagGroup.test.tsx`.

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
KamilEmeleev: Fixed — dropped the trailing `ref`.

One caveat: this removes the churn only for tags without a slot ref, which was every tag before. When a consumer does pass `slotProps.root.ref`, `mergeProps` still merges it with `rowProps.ref` and `mergeRefs` rebuilds the closure each render — unavoidable without memoising a ref that consumers usually recreate inline anyway. TagGroup keeps its trailing `ref` as you noted, since `useTag` returns none.

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
KamilEmeleev: Fixed — though moving the block back to last wasn't enough on its own.

`mergeProps` keeps the earlier value when the later one is `undefined` (`result[key] = b !== undefined ? b : a`), so `'data-selected': isSelected || undefined` still fell through to whatever `slotProps.root` had set. The interaction-state attributes are now assigned after the merge via object spread, where `undefined` genuinely drops the attribute. Covered by a new test in both files.

packages/components/src/components/TagGroup/components/Tag/Tag.tsx
KamilEmeleev: Unified — `slotProps.root` now wins `style` in both wrappers, matching the base `Tag`'s slot contract (the slot is the more specific escape hatch).

That flips TagList's previous direction, which is a deliberate behaviour change and is listed in the PR description. Added a test that sets `style` on both the item and the slot, as you suggested.

packages/components/src/components/Select/components/Tag/Tag.tsx
KamilEmeleev: Checked this one against `main` and I don't think it's a regression.

TagList's own `TagItem.module.css` already read `var(--kbq-tag-bg-color, var(--tag-bg-color))` before this PR, with a `.disabled` rule that set only the private variables — so "the public override beats the disabled state" is long-standing behaviour of TagList and of the standalone `Tag`. `Tag/Tag.module.css` is untouched by this PR. TagGroup and Select converging on that behaviour is precisely what the PR is for, same as the focus-ring change above.

What was genuinely missing is documentation: `TagGroup.mdx` now has a CSS Variables section listing the newly-inherited override points.

packages/components/src/components/TagGroup/components/Tag/Tag.tsx
KamilEmeleev: Intentional — the tags in both wrappers are grid cells, so this is what puts them on the shared markup rather than two hand-rolled copies.

Agreed on both consequences you list, and neither is covered by the test suite. Before/after screenshots and a VoiceOver/NVDA pass are still outstanding — we'll get them into the PR description before this merges rather than asking you to take the layout delta on trust.

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
KamilEmeleev: Fixed.

packages/components/src/components/TagGroup/TagGroup.test.tsx
KamilEmeleev: Added a `removable tags` suite to `TagGroup.test.tsx`: a remove button per tag, removal on press, `tabindex="-1"`, en-US / ru-RU `aria-label`, and a disabled button on a disabled tag that doesn't fire `onRemove`.

You're right about `should be disabled` — it asserted nothing, and `disabledKeys` wasn't even reaching the collection in the fixture (React stringifies `key`, so it needed `'2'`, not `2`). It now passes `disabledKeys` and checks `aria-disabled` / `data-disabled`.

Slot pass-through for `icon`, `content` and `removeIcon` is covered in both files now.

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
KamilEmeleev: Unified to `mergeProps({ tabIndex: -1 }, removeButtonPropsAria, slotProps?.removeIcon)` in both wrappers — slot last, matching the base `Tag`'s contract. `tabIndex` sits in the defaults because `useTag` doesn't set one, while `useTagListItem` does.

This does widen TagList's slot surface: a consumer can now reach `aria-label`, `id` and `onPress` on the remove button. That's deliberate — picking the other order would have made `slotProps.removeIcon` almost decorative — and it's listed in the PR description.

packages/components/src/components/Tag/Tag.tsx
KamilEmeleev: Good catch, and we're keeping it overridable — that's the deliberate choice.

Added a test in both wrappers that replaces the glyph through `slotProps.removeIcon.children`, and it's listed in the PR description as a behaviour change for `TagList.Tag` and TagGroup's `Tag`.

packages/components/src/components/Select/components/Tag/Tag.tsx
KamilEmeleev: Agreed the precedence is inverted, but `Select` is deprecated (`Select.tsx:302`, superseded by `SelectNext`), so we're not changing it here.

It's also unreachable in practice, as you note — both call sites pass only `className` / `isDisabled` / `variant` / `onRemove`. The equivalent order is already correct in the shared `Tag`.

packages/components/src/components/Select/components/Tag/Tag.tsx
KamilEmeleev: Same reasoning — `Select` is deprecated, so the duplicate `utils.ts` / `intl.json` stay until the component is removed. The divergence risk you describe (a new variant reaching TagGroup, TagList and SelectNext but not Select) doesn't outlive the component.

On the bigger question: rendering `<Tag>` there with `slotProps={{ removeIcon: { as: 'div', tabIndex: undefined, onPress: onRemove } }}` would work, but it changes Select's DOM (the `.body` wrapper, `allowsRemoving` gating a button that currently always renders), and this PR deliberately doesn't touch deprecated code.

The only thing this PR changes in `Select` is the stylesheet import, and only because the file it pointed at was deleted.

packages/components/src/components/TagList/TagList.test.tsx
KamilEmeleev: Fixed — both focus assertions are wrapped in `waitFor` now, in `TagList.test.tsx` and in the TagGroup equivalent.

Also took the two nits: `toHaveBeenCalledTimes(1)` is now a plain `toHaveBeenCalled()`, and the `ref.current === null` assertion after `unmount()` is gone — you're right that the `toBe(first)` above is what actually catches a broken ref merge.

packages/components/src/components/TagGroup/components/Tag/Tag.tsx
KamilEmeleev: Fixed — both wrappers now use the three-element tuple.

Left `TagRemoveButtonProps = IconButtonProps<ElementType>` as it is: it's the shared public slot type, and the polymorphic parameter is what lets `as` through for consumers that need a non-`button` element. Happy to narrow it in a follow-up if you'd rather have the stricter shape.

lskramarov (approved):

PR #474, opened 2026-09-08
test: remove console warnings from unit tests (DS-5431)

KamilEmeleev:
## Summary by CodeRabbit

- **Tests**
  - Improved test reliability for keyboard focus, asynchronous highlighting, accessibility labels, resizing, and keyboard shortcuts.
  - Tests now detect unexpected console warnings and errors while allowing known browser-environment messages.
  - Added coverage for validation warnings and improved console mock cleanup.

- **Documentation**
  - Added contributor guidance for handling expected and unexpected console output in tests.

- **Chores**
  - Standardized test configuration across packages and centralized common test setup.


artembelik (approved):

packages/components/src/components/Table/Table.test.tsx
```diff
@@ -293,35 +293,39 @@ describe('Table', () => {
...
-        </Table.Header>
-        <Table.Body items={rows}>
-          {(item) => (
-            <Table.Row>
-              <Table.Cell>{item.name}</Table.Cell>
-            </Table.Row>
-          )}
-        </Table.Body>
-      </Table>
+      <TableContainer>
```
lskramarov: **`TableContainer` reports width `0`, so this silences the warning by accident — and `?? 300` can never fall back**

`TableContainer` takes its width from `useElementSize` → `useResizeObserver`, whose initial state is `width: 0` ([useResizeObserver.ts:10](https://github.com/koobiq/react-components/blob/main/packages/core/src/hooks/useResizeObserver/useResizeObserver.ts#L10)) and which never updates under jsdom — `ResizeObserver` is undefined there and only `matchMedia` is mocked in `setupTests.ts`.

So `tableContainerWidth === 0`, not `undefined`. Two consequences:

1. The guard `tableContainerWidth === undefined` (`Table.tsx:176-183`) goes quiet because `0 !== undefined` — not because the table now has a width.
2. `tableWidth: tableContainerWidth ?? 300` (`Table.tsx:198`) evaluates `0 ?? 300` → **`0`**. This test previously ran the resize layout at `tableWidth: 300`; it now runs it at `0`.

This is the repo's only `isResizable` test, and it now exercises a table width unreachable in a browser. It still passes only because the single column has a static `defaultWidth={200}`; switch it to `1fr` and the assertions describe a 0px layout.

The `??` is also a product bug beyond the test: in a real `<TableContainer><Table isResizable>`, the 300 fallback the comment promises ("Matches the width of the table itself") never applies, because the container publishes `0` on first render too — every non-static column renders at 0px until the first `ResizeObserver` callback.

Suggested: change `Table.tsx:198` to `tableContainerWidth || 300` and widen the guard to `if (!tableContainerWidth)`, then stub `ResizeObserver` here (six other test files already do) so this test exercises a real width.

packages/core/src/utils/FileSizeFormatter/FileSizeFormatter.test.ts
```diff
@@ -84,10 +84,19 @@ describe('FileSizeFormatter', () => {
...
     expect(() => formatter.format(1550, { precision: 101 })).not.toThrow();
+
+    expect(warn).toHaveBeenCalledWith(
+      expect.stringContaining('FileSizeFormatter')
+    );
+
+    warn.mockRestore();
   });
 
   it('falls back to the configured precision when a per-call override is non-finite', () => {
```
lskramarov, line 105: **This test also emits `once.warn` — unmocked. It is silent only because the test above primed the dedupe.**

`isPrecisionInRange` is `Number.isInteger(precision) && …` (`FileSizeFormatter.ts:111-114`), so **both** `Number.NaN` (line 107) and `Number.POSITIVE_INFINITY` (line 112) fail it and reach `once.warn(FILE_SIZE_FORMATTER_ERROR_PRECISION_RANGE)` at `FileSizeFormatter.ts:219`.

`once` dedupes on a module-level `Set` keyed by the message string ([logger/src/index.ts:43,48](https://github.com/koobiq/react-components/blob/main/packages/logger/src/index.ts#L43)), and this file never calls `once.clear()`. So this test passes today only because the test you just fixed at line 86 consumed that message first, under its own spy.

With the new guard that latent ordering dependency becomes a CI/dev failure:

- `pnpm vitest run FileSizeFormatter -t 'non-finite'` → dedupe not primed → real `console.warn` → `vitest-fail-on-console` fails a test that never mentions warnings.
- Same for `it.only` while debugging, or `--sequence.shuffle`.
- Reordering breaks the other direction too: `expect(warn).toHaveBeenCalledWith(...)` at line 95 would see zero calls.

Fix: spy in this test as well, or add `beforeEach(() => once.clear())` to the file (`CodeBlock.test.tsx:36` and `ClampedText.test.tsx:118` already do this).

packages/logger/vite.config.ts
```diff
@@ -14,6 +14,7 @@ export default defineConfig({
   test: {
     globals: true,
     environment: 'jsdom',
+    setupFiles: '../../tools/vitest/setupTests.ts',
```
lskramarov: **The guard is inert in this package.**

`vitest-fail-on-console` patches by direct assignment (`console[m] = collector`) in a **root-suite** `beforeEach` registered when the setup file is evaluated. `packages/logger/src/index.test.ts:9-16` then does:

```ts
beforeEach(() => {
  global.console.warn = vi.fn();
  global.console.error = vi.fn();
});
```

in a **child-suite** `beforeEach`, which always runs after the parent's — overwriting the collector on every test. The capture array can never fill.

Concretely: add `console.warn('stray')` to `packages/logger/src/index.ts` and `pnpm vitest run packages/logger` still passes, while the identical mistake in `core`/`primitives`/`components` fails CI. This line reads as coverage but buys none — and it's the one package whose whole job is console output.

Either convert `index.test.ts` to the `vi.spyOn(...).mockImplementation(() => {})` + `mockRestore()` pattern this PR's own CONTRIBUTING section prescribes (it composes with the guard instead of clobbering it), or drop `setupFiles` here with a comment saying why.

CONTRIBUTING.md
```diff
@@ -34,6 +34,29 @@ Our technology stack:
...
+When a warning or error is expected, mock that console method only in the relevant
+test, assert the message, and restore the mock:
+
+```ts
+it('reports invalid input', () => {
+  const warn = vi.spyOn(console, 'warn').mockImplementation(() => {});
+
+  validateInput('invalid');
+
+  expect(warn).toHaveBeenCalledWith(expect.stringContaining('Invalid input'));
```
lskramarov: **This recipe can't be copy-pasted for the repo's own warnings — it fails on arity.**

`@koobiq/logger` forwards rest args: `logger.warn = (message, ...rest) => console.warn(getMessage(message), ...rest)`. Several real call sites pass a second argument:

- `once.warn('[CodeBlock] File content contains illegal characters.', file)` — `useHighlightedCode.ts:168-171`
- `once.warn('[CodeBlock] Failed to highlight the file.', error)` — `:203`
- the language-fallback warns at `:147-150`

For any of those, `expect(warn).toHaveBeenCalledWith(expect.stringContaining('…'))` fails with *"received 2 arguments, expected 1"*, and the contributor concludes the guard is broken.

The repo's existing tests already do this correctly by spelling out both arguments and the `[koobiq] ` prefix — e.g. `CodeBlock.test.tsx:440-442`:

```ts
expect(warnSpy).toHaveBeenCalledWith(
  '[koobiq] [CodeBlock] Unsupported file language: "unknown". Fall back to "text".',
  file
);
```

Two smaller things in the same snippet: `validateInput` doesn't exist anywhere in the repo, so the example can't be run or grepped for; and it's worth noting that messages routed through `@koobiq/logger` arrive prefixed with `[koobiq] `. Modelling the example on a real call site would fix all three.

tools/vitest/setupTests.ts
```diff
@@ -0,0 +1,7 @@
+import failOnConsole from 'vitest-fail-on-console';
+
+// Expected warnings and errors must be mocked and asserted in the test itself.
+failOnConsole({
+  shouldFailOnWarn: true,
+  shouldFailOnError: true,
+});
```
lskramarov, line 14: **The guard makes every `once.warn` test order-dependent, repo-wide.**

`@koobiq/logger`'s `once`/`deprecate` dedupe on a module-level `Set` keyed by the message string (`logger/src/index.ts:43-52`). Whichever test in a file emits a given message **first** is the only one that can observe it; every later test with the same message is silently suppressed. There are 13+ `once.warn` sites across the components package.

Only three test files reset it — `ClampedText.test.tsx:118`, `CodeBlock.test.tsx:36`, `DropdownMenu.test.tsx:115/525/537`. `ClampedList.test.tsx:141`, `Tabs.test.tsx:170` and the new `FileSizeFormatter` assertion all rely on being first.

Concretely: add a `Tabs` test with `orientation: 'vertical'` + `isUnderlined` above `Tabs.test.tsx:169` and **two** things break at once — the new test fails on an unexpected `console.warn`, and the existing `expect(warnSpy).toHaveBeenCalledTimes(1)` at `:186` fails because the message was already consumed.

Since this file is now the single place that defines console policy, resetting the dedupe here would make the whole rule deterministic:

```ts
import { once } from '@koobiq/logger';
import failOnConsole from 'vitest-fail-on-console';

beforeEach(() => {
  once.clear();
});
```

Without it, the CONTRIBUTING rule *"mock that console method only in the relevant test"* is unsound — "the relevant test" means "whichever happens to run first".

tools/vitest/setupTests.ts
```diff
@@ -0,0 +1,7 @@
+import failOnConsole from 'vitest-fail-on-console';
+
+// Expected warnings and errors must be mocked and asserted in the test itself.
+failOnConsole({
```
lskramarov, line 7: **No escape hatch for console output the project doesn't own — and both flags are already the library defaults.**

`shouldFailOnWarn` and `shouldFailOnError` both default to `true` in `vitest-fail-on-console`, so this options object reads as deliberate policy while being equivalent to `failOnConsole()`. What it *omits* is `silenceMessage`.

jsdom forwards its `"Not implemented: …"` messages to `console.error`; React 19 and React Aria emit dev warnings the repo can't fix at the source. The codebase already hand-patches this class per file — `CodeBlock.test.tsx:59-62` defines `HTMLElement.prototype.scrollTo` with the comment *"jsdom doesn't implement scrolling"*, and `Tabs.test.tsx:227-241` does the same — while production still calls `el.scrollTo(...)` in `Tabs.tsx`, `CodeBlock.tsx` and `useScrollPosition.ts` (which has no test file yet).

AGENTS.md notes React Aria packages are *"pinned to exact versions and updated together"*. One such bump that adds a dev warning turns every affected file red with no source change, and CONTRIBUTING.md:58 explicitly forbids the alternative (*"Do not silence warnings globally"*), so there's no sanctioned path.

A documented `silenceMessage` allowlist for known third-party/jsdom patterns — with a comment and a link per entry — gives the escape hatch a reviewable home. And either drop the two default-valued flags or note in the comment that they're pinned on purpose against an upstream default flip.

CONTRIBUTING.md
```diff
@@ -34,6 +34,29 @@ Our technology stack:
 - `pnpm run lint` — will lint the code once
 - `pnpm run type-check` — will check the typing of the code once
 
+## Console output in tests
+
+All package tests fail on unexpected `console.warn` and `console.error` calls via
```
lskramarov: **"All package tests" over-claims — two workspace members with real tests are outside the guard.**

`pnpm-workspace.yaml` includes `templates/*`, `templates/nextjs/*`, `templates/vite/*`. Both `templates/nextjs/template` and `templates/vite/template` have a `"test": "vitest run"` script, their own vitest config pointing at a local `./setupTests.ts`, and real tests (`src/app/page.test.tsx`, `src/App.test.tsx`). `vite.config.mts:61` globs `packages/**` only, so the shared setup never loads there.

Combined with the guard being inert in `packages/logger` (see the comment on its config), a contributor reads this line, assumes an `act()` warning would have failed the build, and ships a warning-emitting starter template — the code users copy to begin a project.

```suggestion
Tests under `packages/` fail on unexpected `console.warn` and `console.error` calls via
```

packages/components/src/components/CodeBlock/CodeBlock.test.tsx
```diff
@@ -353,6 +357,10 @@ describe('CodeBlock', () => {
       />
     );
 
+    await waitFor(() =>
+      expect(getCode()).toHaveAttribute('data-language', 'javascript')
```
lskramarov: **This `waitFor` moves four previously-synchronous assertions behind the async highlight.**

Lines 365-372 (`code-block-header` has `custom-header`; `getByRole('region')` has `custom-content` and `minBlockSize: 40px`) used to run immediately after `rerender`, which is what proved the slots are applied on the *same commit* as the tabs → no-tabs switch. They now run only once highlighting resolves, so a regression where those slots depend on highlight-derived state would pass silently.

The first added `waitFor` (line 158) has the opposite property — every assertion after it is negative (`toHaveLength(0)`, `not.toHaveBeenCalled()`), so waiting there strengthens the test. No change needed for that one.

If the goal here is only to flush the pending act update, moving this `waitFor` *after* the four synchronous assertions keeps both properties.

pnpm-lock.yaml
```diff
@@ -147,7 +147,7 @@ importers:
         version: 9.1.2(eslint@9.39.5(jiti@2.6.1))
       eslint-plugin-import:
         specifier: ^2.32.0
-        version: 2.32.0(eslint-import-resolver-typescript@3.10.1)(eslint@9.39.5(jiti@2.6.1))
+        version: 2.32.0(@typescript-eslint/parser@8.66.0(eslint@9.39.5(jiti@2.6.1))(typescript@6.0.3))(eslint-import-resolver-typescript@3.10.1)(eslint@9.39.5(jiti@2.6.1))
```
lskramarov: **The lockfile carries dependency-graph churn unrelated to this PR.**

Beyond the two `vitest-fail-on-console` stanzas, this diff also:

- re-keys `eslint-plugin-import@2.32.0` with `@typescript-eslint/parser` in its peer id and gives it a **new** `optionalDependencies: '@typescript-eslint/parser'` entry it didn't have on `main` (lines 150, 11460, 11471), with `eslint-config-next`, `eslint-import-resolver-typescript` and `eslint-module-utils` re-keyed to match (575, 11380, 11420, 11434);
- flips `@vitest/coverage-v8@4.1.9`'s `vitest` peer from `(jsdom@29.1.1)(vite@8.2.1…)` to `(jsdom@25.0.1)(vite@7.3.6…)`;
- rewrites the `git-raw-commits` / `git-semver-tags` deprecation strings (4501/4507/4517) — a sign of a full re-resolve rather than a targeted add.

`import/*` rules resolving through the TS parser instead of the node resolver can start or stop reporting on `.ts`/`.tsx`, and CI runs `lint:js --max-warnings=0`; coverage numbers can move for reasons unrelated to the change. If this is unintentional, re-running `pnpm add -Dw vitest-fail-on-console` from a clean `main` lockfile should reduce the diff to the two new stanzas.

packages/components/vite.config.ts
```diff
@@ -17,7 +17,7 @@ export default defineConfig({
   test: {
     globals: true,
     environment: 'jsdom',
-    setupFiles: './setupTests.ts',
+    setupFiles: ['../../tools/vitest/setupTests.ts', './setupTests.ts'],
```
lskramarov: **This creates the right home for shared setup and then stops one step short.**

`packages/{components,core,primitives}/setupTests.ts` are byte-identical — 15 lines: the `@testing-library/jest-dom/vitest` import plus the `global.matchMedia` mock. `@testing-library/jest-dom` is a **root** devDependency, so it already resolves from `tools/`, and `packages/logger` now loads the shared file and already sets `environment: 'jsdom'`. Folding both lines into `tools/vitest/setupTests.ts` and deleting the three local files collapses `setupFiles` back to a single string in all four configs.

As written, the next jsdom polyfill has to be pasted into three files or it silently applies to only some packages — `ResizeObserver` is already stubbed ad hoc in six separate test files.

The `'../../tools/vitest/setupTests.ts'` literal is also depth-dependent and unenforced. `vite.config.mts:61` already globs `packages/**/vitest.config.ts` (matching nothing today), so a new or nested package becomes a Vitest project with **no** guard — and a missing guard produces no CI signal, so it rots undetected.

The repo has the precedent: root `vite.config.mts` exports `css`, imported at line 11 of this file. Exporting a shared `test` object (or a `defineProject` helper) from `tools/vitest/` the same way would make the wiring self-enforcing.

packages/components/src/components/ClampedList/ClampedList.test.tsx
```diff
@@ -251,7 +251,10 @@ describe('ClampedList', () => {
     expect(getRenderedItems()).toHaveLength(17);
     expect(trigger).toHaveAttribute('aria-expanded', 'true');
 
-    trigger.focus();
+    await user.tab();
```
lskramarov: **`user.tab()` couples a collapse-behaviour test to document tab order — and it's a third idiom for the same fix.**

`ClampedList` renders `<div {...contentProps}>{children(state)}</div>` **before** the trigger, so "the trigger is the first tabbable element" only holds because this test's `renderItems` emits plain `<li>` text. The component's own stories render `<Link href>` inside items, and this same file renders a focusable `List` in content at line 508. Add any focusable element to the fixture and a test titled *"supports defaultExpanded and uncontrolled keyboard toggling"* fails at `expect(getRenderedItems()).toHaveLength(10)` for a tab-order reason. (The `toHaveFocus()` assertion isn't noise — it exists to guard the guess — but it only exists because `tab()` was chosen.)

This PR fixes the identical `element.focus()` act warning with `act(() => el.focus())` in `FileUpload.test.tsx:1060/1077` and `Table.test.tsx:328`, which also matches the existing repo idiom (`Tabs.test.tsx:855`, `TreeSelect.test.tsx:410/793`). Using it here too is one line instead of three, and leaves one answer to "how do I focus without an act warning" for the next contributor to grep for.

packages/components/src/components/TimeRange/TimeRange.test.tsx
```diff
@@ -230,7 +230,7 @@ describe('TimeRange', () => {
     it.each([
       ['Button', <Button key="4" isDisabled={false} />],
       ['Link', <Link key="5" isDisabled={false} />],
-      ['Field', <TimeRange.Field key="6" isDisabled={false} />],
+      ['Field', <TimeRange.Field key="6" label="Period" isDisabled={false} />],
```
lskramarov: **Right fix for the fixture; the API underneath still permits the state that warned.**

In `TimeRange/types.ts`, `label` (`:97`), `aria-label` (`:115`) and `aria-labelledby` (`:116`) are all independently optional. `TimeRangeField.tsx` forwards them to `useField`, and React Aria's `useLabel` emits the generic *"If you do not provide a visible label, you must specify an aria-label or aria-labelledby"* warning — which never names the component.

So `<TimeRange><TimeRange.Field isDisabled /></TimeRange>` type-checks clean and ships a popover trigger with no accessible name (`FormFieldLabel` returns `null` for nil children, so nothing in the DOM names it). The only signal is a dev-mode third-party warning.

Worth considering either requiring a name at the type level — `Base & ({ label: ReactNode } | { 'aria-label': string } | { 'aria-labelledby': string })`, then `pnpm build && pnpm approve-api TimeRange` — or adding a Koobiq-scoped `once.warn` that names the component.

Smaller point: using `aria-label="Period"` in one of these two cases would silence the warning *and* keep the hidden-label rendering shape under test. After this PR, no test renders a `TimeRange.Field` without a visible label.

packages/components/src/components/Sidebar/Sidebar.test.tsx
```diff
@@ -370,7 +376,9 @@ describe('Sidebar', () => {
         cancelable: true,
       });
 
-      window.dispatchEvent(event);
+      act(() => {
```
lskramarov: **`fireEvent(window, event)` does this in one line and is already `act`-wrapped.**

`fireEvent` accepts a caller-built event object, so it keeps the `expect(event.defaultPrevented).toBe(true)` assertion at line 384 working (`defaultPrevented` is set synchronously during dispatch) — without the three-line `act` block or the `act` import that reflowed lines 3-9.

This file dispatches keyboard events 12 other times via `fireEvent.keyDown(window, …)` (291, 312, 325, 341, 344, 353, 354, 370, 394, 410, 433, 449, 462); line 380 is the only raw `dispatchEvent`, and it exists solely so the test can hold onto its own event object. As written, a reader has to work out why this one dispatch is special.

packages/primitives/vite.config.ts
```diff
@@ -15,7 +15,7 @@ export default defineConfig({
   test: {
     globals: true,
     environment: 'jsdom',
-    setupFiles: './setupTests.ts',
+    setupFiles: ['../../tools/vitest/setupTests.ts', './setupTests.ts'],
```
lskramarov: **Worth deleting `vitest.workspace.ts` in this PR** (commenting here since that file isn't in the diff).

This PR rewires `setupFiles` in all four project configs and leaves behind `vitest.workspace.ts`:

```ts
export default ['packages/**/vite.config.ts'];
```

AGENTS.md already documents it as *"Legacy; Vitest 4 ignores it (projects come from `vite.config.mts`)"*. But it still **looks** like the authoritative project list while carrying none of the new wiring, so the next person adding a package or a setup file has even odds of editing the file that does nothing. (Vitest 3.2 also errors when both `test.projects` and a workspace file are present, so a bisect or downgrade hits a stale second definition.)

`git rm vitest.workspace.ts` plus dropping the now-pointless `**/vitest.{workspace,projects}.[jt]s?(on)` coverage exclusion in `vite.config.mts` would finish the job this PR starts.

packages/components/src/components/Table/Table.test.tsx
KamilEmeleev: Confirmed, thanks — reverted the wrap. The test renders the resizable table outside a container again, spies on `console.warn` and asserts the message, so `tableWidth` stays at 300 as before instead of silently dropping to 0.

The `?? 300` product bug is real and I'm not fixing it here: it changes runtime behaviour, so it belongs in a `fix(Table)` commit rather than a test-only PR where it would never reach the changelog. Split into a separate change.

packages/core/src/utils/FileSizeFormatter/FileSizeFormatter.test.ts
KamilEmeleev: Right, the ordering dependency was latent. Added a spy to this test too, and both assertions now use the exported `FILE_SIZE_FORMATTER_ERROR_PRECISION_RANGE` with the `[koobiq] ` prefix instead of `stringContaining`. The shared setup also clears the dedupe cache per test now, so neither test relies on running first.

packages/logger/vite.config.ts
KamilEmeleev: Good catch. Converted `index.test.ts` to `vi.spyOn(console, type).mockImplementation(() => {})` with `vi.restoreAllMocks()` in `afterEach`, which composes with the collector instead of overwriting it. Verified the guard is live in this package.

CONTRIBUTING.md
KamilEmeleev: All three points taken. Rewrote the example on the real `CodeBlock` call site — both arguments spelled out, `[koobiq] ` prefix included — and added a sentence saying the logger forwards rest arguments, so `expect.stringContaining(...)` alone fails on the argument count.

tools/vitest/setupTests.ts
KamilEmeleev: Added `beforeEach(() => once.clear())` here. One wrinkle: `@koobiq/logger` had no `paths` entry, so it did not resolve from `packages/logger` itself — added the alias alongside the other workspace packages. Checked that the logger still bundles as external.

tools/vitest/setupTests.ts
KamilEmeleev: Added `silenceMessage` with one documented entry (jsdom `Error: Not implemented:`, with the upstream link) and a comment stating the bar for new entries. Kept both flags with a note that they are pinned on purpose against an upstream default flip. `CONTRIBUTING.md` now points at this allowlist as the sanctioned path.

CONTRIBUTING.md
KamilEmeleev: Applied the suggestion.

packages/components/src/components/CodeBlock/CodeBlock.test.tsx
KamilEmeleev: Moved it below the four assertions, so they run on the same commit as before and the `waitFor` only flushes the pending update. Left the one at line 158 untouched, as you noted.

pnpm-lock.yaml
KamilEmeleev: Unintentional. Rebuilt from `main`'s lockfile with `pnpm install --lockfile-only`; the diff is now the two `vitest-fail-on-console` stanzas plus three refreshed deprecation strings. The `eslint-plugin-import` re-keying and the `coverage-v8` peer flip are gone.

packages/components/vite.config.ts
KamilEmeleev: Agreed. The three identical `setupTests.ts` files are folded into `tools/vitest/setupTests.ts` and deleted. Added `tools/vitest/config.ts` exporting a shared `test` object, so all four configs are now just `test,` and the path literal exists once.

packages/components/src/components/ClampedList/ClampedList.test.tsx
KamilEmeleev: Switched to `act(() => trigger.focus())`, matching `FileUpload`/`Table`/`Tabs`. One idiom, and the test no longer depends on document tab order.

packages/components/src/components/TimeRange/TimeRange.test.tsx
KamilEmeleev: Took the smaller point: one of the two cases now uses `aria-label="Period"`, so the hidden-label shape stays under test.

Not doing the type-level requirement here — it breaks the public API and needs an `approve-api` run, which does not belong in a test PR. Worth its own issue.

packages/components/src/components/Sidebar/Sidebar.test.tsx
KamilEmeleev: Applied. `defaultPrevented` still asserts correctly and the `act` import is gone, so the import block is back to one line.

packages/primitives/vite.config.ts
KamilEmeleev: Removed the file, dropped the now-pointless coverage exclusion in `vite.config.mts`, and updated the repository tree in `AGENTS.md`.

KamilEmeleev:
Thanks for the review — all fourteen addressed. Two deliberately deferred: the `tableContainerWidth ?? 300` fallback and the `TimeRange.Field` type-level name requirement, both because they change product code or the public API and this PR is test-only.

Full suite green: 108 files, 1751 tests. `type-check`, `build`, `check-api` and `lint` clean.

lskramarov (approved):

PR #475, opened 2026-09-08
feat(SelectNext, TreeSelect): add `dependencies` prop (DS-5530)

KamilEmeleev:
## Summary by CodeRabbit

- **New Features**
  - Added optional dependency tracking to Select and TreeSelect so rendered options and tree items refresh when related values change.
  - Added custom selected-tag rendering for Select and TreeSelect.
  - Expanded public component types and configuration options.

- **Documentation**
  - Added guidance and examples for dependency tracking, including nested sections and list values.

- **Tests**
  - Added coverage for dependency-driven updates and custom tag rendering, including nested sections and tree items.


packages/components/src/components/SelectNext/types.ts
```diff
@@ -149,6 +149,12 @@ export type SelectNextProps<
     defaultInputValue?: string;
     /** Handler that is called when the Select search input value changes. */
     onInputChange?: (value: string) => void;
+    /**
+     * Values the rendered items depend on. The collection caches an item by its
+     * object identity, so a value used inside the render function — a search
+     * query, for instance — has to be listed here for the items to re-render.
+     */
+    dependencies?: ReadonlyArray<unknown>;
```
lskramarov, line 159: **correctness** — `dependencies` never reaches `Select.Section` children.

I ran this against the branch:

```tsx
<Select items={sections} dependencies={[q]}>
  {(s) => (
    <Select.Section id={s.id} title={s.title} items={s.children}>
      {(item) => <Select.Item id={item.id}>{`${item.name}-${q}`}</Select.Item>}
    </Select.Section>
  )}
</Select>
```

Changing `q` from `'a'` to `'b'` leaves the option text at `one-a`.

Root cause: `Select.Section` is `createBranchComponent(SectionNode, SelectSectionInner)` (`SelectSection.tsx:70`). Its built-in `useCollectionChildren` passes only `props.dependencies` to `useCachedChildren` — unlike RAC's `<Collection>`, it does not read `CollectionContext`, so the root-level value is dropped. Adding `dependencies={[q]}` on the section makes it pass.

This is exactly the pattern the `Section` story documents (`Select.stories.tsx:185`). Either forward it — render the section's dynamic children through `<Collection items={...} dependencies={...}>`, the way `DropdownMenuSection.tsx:40` already does — or spell the caveat out in this JSDoc.

packages/components/src/components/SelectNext/__tests__/Select.test.tsx
```diff
@@ -1204,4 +1204,33 @@ describe('Select', () => {
...
+        dependencies={[suffix]}
+        defaultOpen
+      >
+        {(item) => (
+          <Select.Item id={item.id}>{`${item.name}-${suffix}`}</Select.Item>
+        )}
+      </Select>
+    );
+
+    it('should re-render the options when a dependency changes', () => {
```
lskramarov, line 1268: **test-coverage** — the sectioned shape, which is the one that fails, is untested.

This test covers a flat top-level list only. `Select.Section` with `items` + a render function is a first-class documented API (`Select.stories.tsx:156`, `Select.mdx` → `### Sections`) and it is where `dependencies` is silently ignored (see my comment on `types.ts:157`). Worth adding a case here so the suite doesn't report the feature as working across the whole API:

```tsx
it('should re-render options inside a section when a dependency changes', () => {
  // <Select items={sections} dependencies={[suffix]}> with a
  // <Select.Section items={section.children}> render function
});
```

tools/public_api_guard/components/TreeSelect.api.md
```diff
@@ -113,6 +113,7 @@ export type TreeSelectProps<T extends object, M extends SelectionMode_2 = 'singl
     inputValue?: string;
     defaultInputValue?: string;
     onInputChange?: (value: string) => void;
+    dependencies?: ReadonlyArray<unknown>;
```
lskramarov, line 117: **conventions** — `TreeSelect.api.md` was regenerated here, but `SelectNext.api.md` was not.

`SelectNext/types.ts` gained the same public prop and its report is unchanged. The reason it doesn't fail CI is itself the problem: `tools/api-extractor/config.json` lists `"Select"` (the deprecated component) but not `"SelectNext"`, so `tools/public_api_guard/components/SelectNext.api.md` is orphaned — `pnpm approve-api SelectNext` exits with `Unknown target`, and `check-api` never validates it. It's already stale (missing `renderTag`, `minOptionsThreshold`, `isReadOnly`), and this PR widens the drift.

AGENTS.md: *"Any change to an existing component's exported types/signatures needs the same `pnpm build && pnpm approve-api <Name>` step."*

Adding `"SelectNext"` to the `components` array and regenerating would bring it back under the guard.

packages/components/src/components/TreeSelect/types.ts
```diff
@@ -119,6 +119,12 @@ export type TreeSelectProps<
   defaultInputValue?: string;
   /** Handler called when the search query changes. */
   onInputChange?: (value: string) => void;
+  /**
+   * Values the rendered items depend on. The collection caches an item by its
+   * object identity, so a value used inside the render function — a search
+   * query, for instance — has to be listed here for the items to re-render.
+   */
+  dependencies?: ReadonlyArray<unknown>;
```
lskramarov, line 133: **correctness** — `TreeSelectProps` now exposes two `dependencies`, and only this one works.

`slotProps.tree` is `Omit<AriaTreeProps<T>, 'children' | 'items'>` (line 141), and `AriaTreeProps` is RAC's `TreeProps`, which extends `CollectionProps<T>` — so `slotProps.tree.dependencies` type-checks. It does nothing: `slotProps.tree` is merged into `innerTreeProps` and handed to `TreeInner`, which only calls `useTree`, `useRenderProps` and `filterDOMProps` and never builds a collection.

Confirmed on this branch — `slotProps={{ tree: { dependencies: [suffix] } }}` leaves the item at `README.md-a` after `suffix` flips to `'b'`. The tree slot is the natural place to reach for (the items *are* the tree's), so a consumer gets silence instead of a type error.

```suggestion
  dependencies?: ReadonlyArray<unknown>;
```

…and add `'dependencies'` to the `Omit` on line 141 so the dead spelling stops type-checking.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
```diff
@@ -830,4 +830,44 @@ describe('TreeSelect', () => {
...
+          >
+            {function renderItem(item: FileNode) {
+              return (
+                <Tree.Item
+                  key={item.id}
+                  textValue={item.title}
+                  data-testid={`item-${item.id}`}
+                >
+                  <Tree.ItemContent>{`${item.title}-${suffix}`}</Tree.ItemContent>
+                  <Collection items={item.children} dependencies={[suffix]}>
```
lskramarov: **test-quality** — this `dependencies` is redundant, and misleading.

RAC's `<Collection>` concatenates the parent collection's dependencies through `CollectionContext`, so a nested collection inherits them. I checked: dropping it here (and adding `defaultExpandedKeys={[1]}` + an assertion on the nested `item-2`) still passes — nested items refresh from the root prop alone. The file's own fixture at line 63 uses the plain form.

It matters because this test is currently the only worked example of the new prop, and it teaches "repeat `dependencies` on every nested collection" — which is wrong for `<Collection>` but genuinely required for `Select.Section` (see my comment on `SelectNext/types.ts:157`). As written the two components look identical when they aren't.

```suggestion
                  <Collection items={item.children}>
```

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
```diff
@@ -830,4 +830,44 @@ describe('TreeSelect', () => {
...
+            }}
+          </TreeSelect>
+        </Provider>
+      );
+    }
+
+    it('should re-render the items when a dependency changes', () => {
+      const { rerender } = render(<DependenciesFixture suffix="a" />);
+
+      expect(screen.getByTestId('item-7')).toHaveTextContent('README.md-a');
```
lskramarov, line 1013: **test-coverage** — only the root-level leaf is asserted.

`item-7` (`README.md`) is top-level. The nested item (`id: 2`, child of `app`) is collapsed by default, never rendered, and never checked — so nested-branch invalidation, which goes through a second `useCachedChildren` inside the nested `<Collection>`, has no coverage.

I verified the assertion below passes today, so this is a free coverage gain rather than a bug: add `defaultExpandedKeys={[1]}` to the fixture and

```tsx
expect(screen.getByTestId('item-2')).toHaveTextContent('Http-a');
// …after rerender
expect(screen.getByTestId('item-2')).toHaveTextContent('Http-b');
```

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
```diff
@@ -830,4 +830,44 @@ describe('TreeSelect', () => {
       expect(screen.getByTestId('item-7')).toBeInTheDocument();
     });
   });
+
+  describe('dependencies', () => {
+    function DependenciesFixture({ suffix }: { suffix: string }) {
```
lskramarov: **simplification** — this duplicates `TreeSelectFixture` (line 27).

Same `Provider`, same `items`, same recursive `renderItem`, same `data-testid` scheme — ~25 lines differing only by the `suffix` interpolation and the `dependencies` prop. Two fixtures now have to move together whenever the shared tree rendering changes, and the copy sits 800 lines below the original where it's easy to miss.

`TreeSelectFixture` already takes `Partial<TreeSelectProps<FileNode, M>>`; giving it an optional `suffix` (empty by default) and reusing it here removes the duplicate.

packages/components/src/components/TreeSelect/types.ts
```diff
@@ -119,6 +119,12 @@ export type TreeSelectProps<
   defaultInputValue?: string;
   /** Handler called when the search query changes. */
   onInputChange?: (value: string) => void;
+  /**
```
lskramarov, line 126: **reuse / altitude** — this JSDoc block and prop are pasted verbatim into `SelectNextProps` too.

One wording, two places today, one more per component later. And the pitfall isn't specific to these two: `FileUpload.tsx:464` renders `<Collection items={items}>{children}</Collection>` with no `dependencies` at all, and `Autocomplete` / `TagAutocomplete` build from `@react-types` prop shapes that never had it. (The `Autocomplete` `HighlightingMatches` story only escapes it because it rebuilds `items` on every render — `Autocomplete.stories.tsx:277`.)

A single `CollectionDependenciesProps` (or re-exporting RAC's `CollectionProps['dependencies']` from core) intersected into each collection component would keep the contract and its documentation in one place, and make the remaining components a one-line change instead of another copy of this comment.

packages/components/src/components/SelectNext/types.ts
```diff
@@ -149,6 +149,12 @@ export type SelectNextProps<
     defaultInputValue?: string;
     /** Handler that is called when the Select search input value changes. */
     onInputChange?: (value: string) => void;
+    /**
```
lskramarov, line 152: **conventions (docs)** — new public prop on two documented components, no MDX section and no story.

AGENTS.md describes the MDX shape as *"`## Props`, then one section per feature with `<Story of={Stories.Feature} />`"*. `dependencies` is the least self-explanatory prop in either API — it exists to work around an invisible identity cache, and per my comment on line 157 it behaves differently for sectioned Selects — so the auto-generated Props table alone leaves consumers to discover the rule from a stale-label bug.

Natural homes: next to `### Searchable` / `### Server search` in `Select.mdx` (line 181) and `### Searchable` in `TreeSelect.mdx` (line 138), with a story that highlights the live query inside each item — which is the case the JSDoc names.

packages/components/src/components/SelectNext/types.ts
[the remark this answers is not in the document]
KamilEmeleev: Regenerated in b4a58346.

One correction: `pnpm check-api` would not have failed. `SelectNext` was missing from the `components` array in `tools/api-extractor/config.json`, so the report was never validated — that was the actual problem (see @lskramarov's comment on `TreeSelect.api.md`). `SelectNext` is now in the config and the report is regenerated, so from here on a stale report does fail CI.

KamilEmeleev: Confirmed and fixed in b66e3081.

Traced it to the same place you did: `createBranchComponent` calls `useChildren(props)`, defaulting to `useCollectionChildren`, which forwards only `props.dependencies` — unlike `Collection`, which concatenates `ctx.dependencies` from `CollectionContext` (`@react-aria/collections/CollectionBuilder`, `Collection`).

Rather than documenting the caveat or asking consumers to repeat the prop, I passed the optional third argument of `createBranchComponent` so the section renders its children through `Collection`:

```tsx
const SelectSectionRoot = createBranchComponent(
  SectionNode,
  SelectSectionInner,
  ({ items, children }) => <Collection items={items}>{children}</Collection>
);
```

The section now inherits both `dependencies` and `idScope` from the Select, so `dependencies` only has to be set once, on the root — same as nested `Collection` in TreeSelect. `useCachedChildren` returns static children untouched when `children` is not a function, so the non-dynamic form is unaffected.

packages/components/src/components/SelectNext/__tests__/Select.test.tsx
KamilEmeleev: Added in b66e3081 — `should re-render the options inside a section when a dependency changes`, with `Select.Section` + `items` + a render function, exactly the documented shape.

Checked it is a real regression test: with the `SelectSection` fix stashed it fails, with it applied it passes.

tools/public_api_guard/components/TreeSelect.api.md
KamilEmeleev: Right on all counts — the report was orphaned, unvalidated and stale. Fixed in b4a58346.

Adding `"SelectNext"` to the `components` array turned out not to be enough. API Extractor crashes on the entry point:

```
API Extractor: SelectNext — crashed: Internal Error: Unable to follow symbol for "T_1"
```

Bisected it to `Section: typeof SelectSection` in the compound type. The declaration emitted for `SelectSection` leaks an unresolved type parameter:

```ts
export declare const SelectSection: <T extends object>(props: Omit<
  DetailedHTMLProps<HTMLAttributes<HTMLElement>, HTMLElement>,
  "id" | keyof SectionProps<T_1>   // ← T_1
> & SectionProps<T> & { id?: Key } & RefAttributes<HTMLElement>) => ReactElement | null;
```

This reproduces on `main`, so it is pre-existing — presumably why the component was left out of the config in the first place. Annotating the export with an explicit `SelectSectionComponent` type keeps the inferred form out of the declaration output:

```ts
export declare const SelectSection: SelectSectionComponent;
```

With that, the report generates and `pnpm check-api` exits 0. The regenerated report picks up everything that had drifted: `renderTag`, `minOptionsThreshold`, `isReadOnly`, and `dependencies`. It also switches from the old prettier-formatted layout to the raw API Extractor one, which is what every other report in `tools/public_api_guard/` uses (the directory is in `.prettierignore`) — hence the large diff.

packages/components/src/components/TreeSelect/types.ts
KamilEmeleev: Confirmed — `TreeInner` only calls `useTree`, `useRenderProps` and `filterDOMProps`, so the slot spelling is dead. Added `'dependencies'` to the `Omit` in f78a02bc:

```ts
tree?: Omit<AriaTreeProps<T>, 'children' | 'items' | 'dependencies'> &
  DataAttributeProps;
```

Kept the explicit root prop rather than relying on inheritance. Worth noting for the record: `TreeSelectProps` already accepted `dependencies` before this PR — `AriaTreeSelectProps` is `TreeSelectStateOptions`, which extends `TreeProps<T>` → `CollectionProps<T>` — and it already worked at runtime, since `TreeSelect.tsx:471` spreads props into `<Collection {...props} />`. So the declaration here is documentation of an existing capability, and the `Omit` makes the root the only place it can be spelled.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
KamilEmeleev: Removed in f78a02bc. Verified the same way you did — `Collection` concatenates `ctx.dependencies`, so the nested one inherits from the root and the test still passes without it.

The point about the two components looking identical when they aren't is now moot in the other direction: `Select.Section` inherits too, after b66e3081. So the rule is the same everywhere — set `dependencies` once, on the root.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
KamilEmeleev: Added in f78a02bc — `defaultExpandedKeys={[1]}` on the fixture plus assertions on `item-2` before and after the rerender, so the nested `useCachedChildren` is covered.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
KamilEmeleev: Done in f78a02bc. `TreeSelectFixture` now takes an optional `suffix` (empty by default, so every existing test is unchanged) and the duplicate is gone:

```tsx
const renderWithSuffix = (suffix: string) => (
  <TreeSelectFixture
    suffix={suffix}
    dependencies={[suffix]}
    defaultExpandedKeys={[1]}
    defaultOpen
  />
);
```

packages/components/src/components/TreeSelect/types.ts
KamilEmeleev: Agreed on the direction, but I'd like to do it separately rather than here.

Deduplicating the JSDoc across two components is the small half of the value; the real payoff is giving the prop to `FileUpload`, `Autocomplete` and `TagAutocomplete`, which is a wider change than DS-5530. A shared `CollectionDependenciesProps` also has to be exported to avoid an `ae-forgotten-export` warning, which adds a new public type to both api reports for what is currently a documentation concern.

Happy to open a follow-up covering all collection components at once — the shared type lands there together with its first real reuse.

One note that changes the shape of that task: after b66e3081 the contract is uniform. `Select.Section` inherits `dependencies` from the root the same way nested `Collection` does, so the shared documentation can state one rule — set it on the root — instead of carrying a per-component caveat.

packages/components/src/components/SelectNext/types.ts
KamilEmeleev: Added in 4aecac59 — a `### Dependencies` section plus a `Dependencies` story in both `Select.mdx` (after `### Searchable`) and `TreeSelect.mdx` (same place).

The story is the case the JSDoc names: a controlled `inputValue` with `Highlight` rendering the live query inside each item, which is stale without `dependencies`.

The caveat you asked to spell out is gone rather than documented — after b66e3081 sections inherit the prop, so the docs state a single rule for both components: set `dependencies` on the root, nested collections and sections inherit it.

Verified both stories in Storybook; the TreeSelect one highlights nested items correctly, which is the inheritance path in action. Also bumped the `date:` tag on both stories to `2026-09-09`.

packages/components/src/components/SelectNext/types.ts
lskramarov, line 159: **`dependencies` must have a constant length — worth saying in the doc comment**

This value is spliced straight into React `useMemo` dependency arrays by React Aria: `useCachedChildren` does `useMemo(() => new WeakMap(), dependencies)`, and `Collection` does `useMemo(..., [idScope, ...dependencies])`. So the array has the same constraint as a hook dependency list — its length must not change between renders.

Nothing here, in the identical JSDoc at `TreeSelect/types.ts:131`, or in the two new MDX sections says so. A consumer who follows the docs and writes `dependencies={activeFilters}` will, the moment a second filter is added, get React's *"The final argument passed to useMemo changed size between renders"* warning — and, because React only compares up to the shorter array, the item cache is never invalidated. The options keep rendering the stale value with no error.

One extra sentence in the JSDoc (and the MDX) would close it.

packages/components/src/components/SelectNext/components/SelectSection/SelectSection.tsx
```diff
@@ -69,7 +74,15 @@ function SelectSectionInner<T extends object>(
...
 
-export const SelectSection = createBranchComponent(
+const SelectSectionRoot = createBranchComponent(
   SectionNode,
-  SelectSectionInner
+  SelectSectionInner,
+  // Render the children through `Collection` rather than the built-in
+  // `useCollectionChildren`, so the section inherits `dependencies` from the
+  // Select it is rendered in.
+  ({ items, children }) => <Collection items={items}>{children}</Collection>
```
lskramarov: **This drops `dependencies` and `idScope` from the section's own props**

The default third argument is `useCollectionChildren(props)` = `useCachedChildren({ ...props, addIdAndValue: true })`, which forwarded `props.dependencies` and `props.idScope`. The replacement destructures only `items` and `children`, so both are silently dropped — no type error, no warning.

Today that mostly means `Select.Section` can never gain the per-section override that the sibling already supports (`DropdownMenuSection.tsx:40` renders `<Collection items={items} dependencies={dependencies}>`), and that a JS consumer passing `dependencies` on the section gets a silent no-op. Forwarding them costs one prop each, and RAC's `Collection` still concatenates the inherited context values on top, so inheritance from the root is unaffected:

```suggestion
  ({ items, children, dependencies, idScope }) => (
    <Collection items={items} dependencies={dependencies} idScope={idScope}>
      {children}
    </Collection>
  )
```

(If the intent is that the root is the only place `dependencies` may be set, that is worth stating in the comment above — right now the comment reads as if nothing was given up.)

packages/components/src/components/SelectNext/Select.stories.tsx
```diff
@@ -543,6 +544,31 @@ export const Searchable: Story = {
...
 
+export const Dependencies: Story = {
+  render: function Render() {
+    const [inputValue, setInputValue] = useState('');
+
+    return (
+      <Select
+        items={options}
+        label="Attack type"
+        inputValue={inputValue}
```
lskramarov: **The controlled `inputValue` is never cleared, so the demo keeps a stale query**

`SelectInner` deliberately skips its reset effect when the search is controlled (`Select.tsx:120-138` returns early on `inputValueProp !== undefined`), so this story has to clear it itself.

On the published docs page: type `dd`, pick *DDoS*, the popover closes — reopen it and the search box still reads `dd` and the list is still narrowed to one option. The uncontrolled `Searchable` story directly above resets, so the two examples behave differently for no reason the prose explains.

```suggestion
        inputValue={inputValue}
        onOpenChange={(isOpen) => {
          if (!isOpen) setInputValue('');
        }}
```

Same thing in the TreeSelect `Dependencies` story (`TreeSelect.stories.tsx:612`).

packages/components/src/components/SelectNext/Select.mdx
```diff
@@ -186,6 +186,18 @@ with `defaultInputValue`, and the `onInputChange` callback is called whenever th
 
 <Story of={Stories.Searchable} />
 
+### Dependencies
+
+The collection caches a rendered item by the identity of its data object, so an item is not
+re-rendered when a value used inside the render function changes. List such values in
+`dependencies` to invalidate that cache — for example the search query when the options
```
lskramarov, line 193: **The search query is the one dependency the component already owns**

Both new docs sections use the search query as the motivating — and only — example. But `SelectInner` holds `inputValue` itself via `useControlledState`, and `TreeSelectInner` does the same; the collection is built one level above, in `SelectRender` / `TreeSelectRender`, with `<Collection {...props} />`.

So for the canonical case the API asks the consumer to hand back a value the component already has, and anyone who combines `isSearchable` with per-item highlighting and forgets `dependencies={[inputValue]}` gets silently stale, unhighlighted options.

Worth considering lifting the search value so the root `Collection` gets `dependencies={[...(props.dependencies ?? []), searchValue]}` whenever `isSearchable` is set. That makes the common case correct by construction and leaves `dependencies` for genuinely external values — which is what the prop is actually for.

Not blocking; the prop as specified is still useful either way.

packages/components/src/components/TreeSelect/types.ts
```diff
@@ -136,7 +142,8 @@ export type TreeSelectProps<
     control?: FormFieldSelectProps;
     popover?: PopoverProps;
     dropdownFooter?: DropdownFooterProps & DataAttributeProps;
-    tree?: Omit<AriaTreeProps<T>, 'children' | 'items'> & DataAttributeProps;
+    tree?: Omit<AriaTreeProps<T>, 'children' | 'items' | 'dependencies'> &
```
lskramarov, line 147: **`idScope` is dead here for exactly the same reason**

`dependencies` and `idScope` both come from the `CollectionProps` base that RAC's `TreeProps` extends, and `slotProps.tree` is forwarded to `TreeInner`, which never builds a collection — it runs the props through `filterDOMProps` (`TreeInner.tsx:337`) and drops anything that is not a DOM prop.

So `slotProps={{ tree: { idScope: 'x' } }}` still type-checks and still does nothing: the same trap this change closes for `dependencies`.

```suggestion
    tree?: Omit<
      AriaTreeProps<T>,
      'children' | 'items' | 'dependencies' | 'idScope'
    > &
```

packages/components/src/components/SelectNext/Select.mdx
```diff
@@ -186,6 +186,18 @@ with `defaultInputValue`, and the `onInputChange` callback is called whenever th
...
+highlight it.
+
+The same applies to `Select.Section`: it inherits `dependencies` from the Select, so the
```
lskramarov, line 199: **This paragraph has no example behind it**

`Stories.Dependencies`, rendered three lines below, contains no sections — so the inheritance this promises is stated but never shown, and the docs page never exercises the `SelectSection` code path that commit b66e3081 exists to fix. A regression there would surface only in the unit test, not on the docs site or in a visual review.

The section variant already exists as a test fixture (`renderSectionsWithSuffix`, `Select.test.tsx:1232`); promoting it to a story — or just adding sections to the existing one — would cover the claim and the code path at once.

packages/components/src/components/SelectNext/Select.stories.tsx
```diff
@@ -543,6 +544,31 @@ export const Searchable: Story = {
   },
 };
 
+export const Dependencies: Story = {
+  render: function Render() {
+    const [inputValue, setInputValue] = useState('');
+
+    return (
+      <Select
+        items={options}
```
lskramarov, line 553: **Story data belongs inside `render` — and here it carries the point**

AGENTS.md, *Storybook Stories*: "Define story data and helpers inside `render` so they appear in the Source panel (the docs page shows the raw text of the story export)."

The Source panel for this story shows `items={options}` with no definition in sight. That matters more than usual here: the surrounding prose explains that the collection caches by *object identity*, and a reader cannot see that `options` is a stable module-level array of stable objects — which is the entire reason `dependencies` is needed.

Same at `TreeSelect.stories.tsx:610`. (The neighbouring stories share the habit, so this is a judgement call on whether to fix it just for the new ones.)

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
```diff
@@ -998,4 +998,27 @@ describe('TreeSelect', () => {
       expect(screen.getByTestId('item-7')).toBeInTheDocument();
     });
   });
+
+  describe('dependencies', () => {
+    const renderWithSuffix = (suffix: string) => (
+      <TreeSelectFixture
```
lskramarov: **This is the only test in the file that skips `renderTreeSelect`**

It has to, because `renderTreeSelect` (line 72) still takes `Partial<TreeSelectProps<FileNode, M>>` and was not widened with the `suffix` prop `TreeSelectFixture` gained at line 28. The result is two render paths in one file: anything later added to the helper (a wrapper, a default prop, cleanup) silently will not apply to the `dependencies` test.

Mirroring the fixture's signature restores one path:

```ts
function renderTreeSelect<M extends SelectionMode = 'single'>(
  props: Partial<TreeSelectProps<FileNode, M>> & { suffix?: string } = {}
) {
  return render(<TreeSelectFixture {...props} />);
}
```

packages/components/src/components/SelectNext/components/SelectSection/SelectSection.tsx
```diff
@@ -69,7 +74,15 @@ function SelectSectionInner<T extends object>(
...
 );
+
+// The type is spelled out: the inferred one leaks an unresolved type parameter
+// into the declaration output, which API Extractor cannot follow.
+export const SelectSection = SelectSectionRoot as SelectSectionComponent;
```
lskramarov, line 98: **displayName is still `SelectSectionInner`**

`createBranchComponent` sets `Result.displayName = render.name`, so React DevTools, Storybook's `subcomponents` entry and any snapshot show `SelectSectionInner` rather than `Select.Section`. AGENTS.md, *Component Anatomy*: "Always set `X.displayName`" — and the sibling does (`DropdownMenuSection.displayName = 'DropdownMenu.Section'`).

Pre-existing, but this change makes it slightly harder to fix later (the cast target has no `displayName`), and now that there is a named `SelectSectionRoot` it is a one-liner before the cast:

```ts
SelectSectionRoot.displayName = 'Select.Section';
```

tools/api-extractor/config.json
```diff
@@ -49,6 +49,7 @@
     "Resizable",
     "SearchInput",
     "Select",
+    "SelectNext",
```
lskramarov, line 52: **Adopting the guard here makes this PR's real API delta unreadable**

Bringing `SelectNext` under the guard replays every API change accumulated while the report sat unguarded, so `SelectNext.api.md` moves 325 lines for a one-prop feature. Besides `dependencies`, the regenerated report picks up `renderTag`, `isReadOnly`, `minOptionsThreshold`, `SelectNextItemProps`, `SelectNextSectionProps`, `SelectNextTagProps`, `SelectNextDividerProps`, a new `isReadOnly` in the `renderValue` state object, and an LF→CRLF rewrite of the whole file.

A reviewer approving that diff cannot tell which entries this PR intends. Splitting it into its own PR is the clean version — `main` squashes to one commit per PR, so keeping it as commit b4a58346 does not preserve the separation.

Good change to make, though; the report was drifting.

lskramarov (approved):

packages/components/src/components/SelectNext/types.ts
KamilEmeleev: Added in c39a4afb: one sentence in both JSDoc blocks and both MDX sections. The array must keep the same length between renders, and a list is passed wrapped: `dependencies={[filters]}`.

packages/components/src/components/SelectNext/Select.stories.tsx
KamilEmeleev: Fixed in c39a4afb, slightly differently: both stories no longer pass `inputValue`. The search stays uncontrolled, so the component clears it on close, like in the `Searchable` story. The story only mirrors the value through `onInputChange` for `dependencies` and `Highlight`. Checked in Storybook: type `dd`, pick DDoS, reopen — empty search, full list, no highlight. Same for TreeSelect.

packages/components/src/components/SelectNext/components/SelectSection/SelectSection.tsx
KamilEmeleev: Went with the per-section override, like `DropdownMenu.Section` (c39a4afb). Neither `dependencies` nor `idScope` was part of `SelectSectionProps`, so `Select.Section` now has a typed `dependencies` prop, forwarded to its `Collection` on top of the inherited ones. One caveat, now in the JSDoc: it takes effect for a section written out in JSX. A section rendered from the Select's `items` is cached by the Select, so there the value still goes on the root. Covered by a new test. `idScope` isn't forwarded: the section already inherits it from the Select.

packages/components/src/components/TreeSelect/TreeSelect.test.tsx
KamilEmeleev: Done in c39a4afb. `renderTreeSelect` now takes `suffix`, and the test goes through it like the neighbouring `rerender` tests: first render via the helper, `rerender` with the fixture.

packages/components/src/components/SelectNext/components/SelectSection/SelectSection.tsx
KamilEmeleev: Pre-existing and out of scope here. No part of SelectNext or TreeSelect sets `displayName` yet, so I'd rather not fix one in isolation.

packages/components/src/components/SelectNext/Select.stories.tsx
KamilEmeleev: Keeping `options` module-level on purpose. Inside `render`, the array would be recreated on every keystroke and every item would re-render anyway. The story would then work without `dependencies`, hiding exactly what it demonstrates. Real data usually has a stable identity, and that's the case the story models.

packages/components/src/components/SelectNext/Select.mdx
KamilEmeleev: Leaving it as prose: it's a one-line rule. The section code path is covered by unit tests for both inherited and per-section `dependencies`, and that's where a regression would be caught.

packages/components/src/components/TreeSelect/types.ts
KamilEmeleev: This one doesn't reproduce. `slotProps.tree` is based on RAC's `TreeProps`, whose `CollectionProps` only adds `children` and `dependencies`. `idScope` lives on the `CollectionProps` from `@react-aria/collections`, which `TreeProps` doesn't extend. So `slotProps={{ tree: { idScope: 'x' } }}` is already a type error — checked with `tsc`.

packages/components/src/components/SelectNext/Select.mdx
KamilEmeleev: Keeping it explicit. Filtering doesn't rebuild the collection today. An automatic dependency would re-render every item on each keystroke in every searchable Select, including ones that don't highlight. And a consumer who highlights already holds the query for `Highlight`, so it's one extra `dependencies={[query]}`.

tools/api-extractor/config.json
KamilEmeleev: Keeping it in this PR. Splitting now means reverting b4a58346 here (force-push is off-limits on this repo) and landing it separately, while the TreeSelect report from the same commit would have to stay. That's more churn than it saves. For the record, everything the report picked up besides `dependencies` already shipped on `main`: `renderTag`, `isReadOnly`, `minOptionsThreshold`, the `SelectNext*Props` aliases, and `isReadOnly` in `renderValue`. The line endings are API Extractor's default CRLF, same as 73 of the 76 reports in `tools/public_api_guard/`. The old file was LF only because it had been prettier-formatted.

PR #476, opened 2026-09-08
docs(Input): add input mask example (DS-5485)

KamilEmeleev:


packages/components/src/components/Input/Input.stories.tsx
```diff
@@ -451,3 +454,148 @@ export const Validation: Story = {
...
+          index === 0 ? [hex, hex] : [':', hex, hex]
+        ),
+        postprocessors: [toUpperCase],
+      }),
+      []
+    );
+
+    const ipv6Options = useMemo(
+      () => ({
+        mask: Array.from({ length: 8 }).flatMap((_, index) =>
```
lskramarov: **IPv6 mask silently rewrites canonical addresses** _(correctness)_

The mask hard-codes eight zero-padded 4-hex-digit groups, so anything written in the canonical [RFC 5952](https://www.rfc-editor.org/rfc/rfc5952) form — compressed `::`, or groups without leading zeros — is silently rewritten into a *different* address instead of being rejected.

Verified by running these exact options through `maskitoTransform` (`@maskito/core@5.4.0`):

| input | result |
| --- | --- |
| `2001:db8:85a3:0:0:8a2e:370:7334` | `2001:db88:5a30:08a2:e370:7334` |
| `2001:db8::1` | `2001:db81` |
| `::1` | `1` |
| `0:0:0:0:0:0:0:1` | `0000:0001` |

Typing is affected too: after `2001:db8` the `:` key is swallowed (`2001:db8:` → `2001:db8`), so a group shorter than 4 digits can never be terminated — the canonical form cannot be typed at all, only the fully expanded one.

A user pastes a valid address, sees a plausible-looking but different one, and `onChange` reports the corrupted value. For a security-product component library that seems worth avoiding in the flagship example — either accept `::` and variable-length groups, or drop the IPv6 field and keep MAC/license key (which both behave correctly).

packages/components/src/components/Input/Input.stories.tsx
```diff
@@ -451,3 +454,148 @@ export const Validation: Story = {
...
+      value,
+      selection,
+    }: {
+      value: string;
+      selection: readonly [number, number];
+    }) => ({ value: value.toUpperCase(), selection });
+
+    const ipv4Options = useMemo(
+      () => ({
+        mask: /^((25[0-5]|2[0-4]\d|[01]?\d?\d)\.){0,3}(25[0-5]|2[0-4]\d|[01]?\d?\d)?$/,
```
lskramarov, line 483: **IPv4 regex mask corrupts pasted addresses** _(correctness)_

Because the mask is a `RegExp`, Maskito can only *drop* characters that break the pattern (`guessValidValueByRegExp`) — it cannot insert fixed separators and it cannot reject a paste. Extra groups and suffixes get merged into the previous octet:

| input | result |
| --- | --- |
| `1.2.3.4.5` | `1.2.3.45` |
| `192.168.1.1:8080` | `192.168.1.180` |

Both produce a valid-looking but wrong address, silently, and `onChange` reports it.

Separately, this is the only field of the five that auto-inserts nothing. Typing `1921680` yields `192` — every digit after the third is swallowed until the user types `.` by hand, which contradicts the `Format: 192.168.0.1` caption and the auto-separating behaviour readers just saw in the MAC / IPv6 / license-key fields right below.

An array mask (`[d, d, d, '.', d, d, d, …]` with a `mask` callback for per-octet range checking) would both insert the dots and reject out-of-range input consistently with the other examples.

packages/components/src/components/Input/Input.stories.tsx
```diff
@@ -451,3 +454,148 @@ export const Validation: Story = {
...
+    const [ipv4, setIpv4] = useState('192.168.0.1');
+    const [mac, setMac] = useState('AA:BB:CC:DD:EE:FF');
+
+    const [ipv6, setIpv6] = useState('2001:0db8:85a3:0000:0000:8a2e:0370:7334');
+
+    const [port, setPort] = useState('8080');
+    const [licenseKey, setLicenseKey] = useState('ABCD-1234-EFGH-5678');
+
+    const handlePortKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
+      if (event.key !== 'ArrowUp' && event.key !== 'ArrowDown') return;
```
lskramarov: **Arrow-key stepper ignores `isReadOnly`** _(correctness)_

`handlePortKeyDown` mutates the value on ArrowUp / ArrowDown without checking `isReadOnly` (or the form-level read-only state), so a read-only field stays editable from the keyboard.

Reachable from this story itself: `{...args}` is spread onto the Port input, so toggling `isReadOnly` in the Controls panel gives a field where typing is correctly blocked by the browser but ArrowUp still fires `setPort` and changes the displayed value. Anyone copying this pattern into a form that toggles read-only inherits the same hole.

```suggestion
    const handlePortKeyDown = (event: KeyboardEvent<HTMLInputElement>) => {
      if (event.currentTarget.readOnly) return;

      if (event.key !== 'ArrowUp' && event.key !== 'ArrowDown') return;
```

packages/components/src/components/Input/Input.stories.tsx
```diff
@@ -451,3 +454,148 @@ export const Validation: Story = {
...
+    const ipv6Options = useMemo(
+      () => ({
+        mask: Array.from({ length: 8 }).flatMap((_, index) =>
+          index === 0 ? [hex, hex, hex, hex] : [':', hex, hex, hex, hex]
+        ),
+      }),
+      []
+    );
+
+    const portOptions = useMemo(
```
lskramarov: **Port field re-implements `InputNumber` / `NumberField`** _(reuse)_

This builds a bounded integer input from `maskitoNumberOptionsGenerator` plus a hand-written ArrowUp/ArrowDown handler — which is exactly what `InputNumber` (React Aria `NumberField`) already provides: `minValue` / `maxValue` / `step`, arrow-key stepping, formatting and spinbutton semantics.

The `<Alert>` added directly above this story ([Input.mdx:130](https://github.com/koobiq/react-components/blob/231beb28d27c957e4012fc3e6b036b7145ad6a8f/packages/components/src/components/Input/Input.mdx#L130)) says:

> For numbers, currencies, and units consider InputNumber first: it formats the value out of the box and doesn't need an external library.

…and then the example immediately demonstrates the opposite. Readers will copy the Maskito + manual-stepper recipe for numeric fields instead of `<InputNumber minValue={0} maxValue={65535} />`.

Suggest dropping the Port field from this story (the remaining four are all genuine mask cases that `InputNumber` can't cover), or keeping it only as a counter-example that points at `InputNumber`.

packages/components/src/components/Input/Input.stories.tsx
```diff
@@ -451,3 +454,148 @@ export const Validation: Story = {
...
+          value={ipv6}
+          onChange={setIpv6}
+          slotProps={{ input: { ref: ipv6Ref } }}
+          isClearable
+          fullWidth
+          {...args}
+        />
+        <Input
+          label="Port"
+          caption="Range: 0-65535. Use ↑/↓ to increase or decrease the value"
```
lskramarov: **Hand-rolled stepper has no spinbutton semantics** _(accessibility)_

The arrow-key stepping is exposed through a plain textbox: no `role="spinbutton"`, no `aria-valuenow` / `aria-valuemin` / `aria-valuemax`. The affordance is discoverable only by reading this caption visually.

A screen-reader user focuses the field and hears a textbox plus the description text — no indication it is a spinner, no current/min/max value, and no announcement when ArrowUp/ArrowDown change the value. In a library built on React Aria specifically for accessibility, this is the snippet readers copy.

`InputNumber` / `NumberField` supplies all of this for free (see the related comment on `portOptions`).

packages/components/src/components/Input/Input.mdx
```diff
@@ -116,6 +117,24 @@ See the example of how to present the component with a password input.
 
 <Story of={Stories.Password} />
 
+## Input mask
+
+Input masks control the format of the value and validate it in real time.
```
lskramarov: **"validate it in real time" overstates what a mask does** _(docs accuracy)_

None of the five examples validate — every mask happily accepts incomplete values, and the story wires up no validation at all:

- `192.168` and `1.2` pass the IPv4 mask
- `AA:B` passes the MAC mask
- `2001:db8` passes the IPv6 mask

All of them are reported through `onChange` as accepted values. A reader taking this sentence at face value ships an IP field with no validation.

Suggest saying that a mask *constrains what can be typed* and pointing at the [Validation](#validation) section below for actually rejecting incomplete input.

pnpm-lock.yaml
```diff
@@ -147,7 +156,7 @@ importers:
         version: 9.1.2(eslint@9.39.5(jiti@2.6.1))
       eslint-plugin-import:
         specifier: ^2.32.0
-        version: 2.32.0(eslint-import-resolver-typescript@3.10.1)(eslint@9.39.5(jiti@2.6.1))
+        version: 2.32.0(@typescript-eslint/parser@8.66.0(eslint@9.39.5(jiti@2.6.1))(typescript@6.0.3))(eslint@9.39.5(jiti@2.6.1))
```
lskramarov: **Unrelated lockfile churn in a docs-only PR** _(hygiene)_

Beyond the three `@maskito/*` additions, this regeneration also re-resolves parts of the toolchain:

- `eslint-plugin-import@2.32.0` flips from the `(eslint-import-resolver-typescript@3.10.1)` variant to an `(@typescript-eslint/parser@8.66.0)` variant, **while the old snapshot block stays in the file** — so the repo now carries two copies of the plugin and of `eslint-module-utils`.
- `@vitest/coverage-v8`'s vitest peer moves from `vitest(jsdom@29.1.1)(vite@8.2.1)` to `vitest(jsdom@25.0.1)(vite@7.3.6)`.

Neither is caused by adding a docs example. They make the diff hard to review and can silently shift lint/coverage behaviour. Suggest regenerating from `main` with only the `@maskito/*` additions, or landing the drift separately as its own `build(deps)` commit.

packages/components/src/components/Input/Input.stories.tsx
```diff
@@ -451,3 +454,148 @@ export const Validation: Story = {
...
+    const ipv4Options = useMemo(
+      () => ({
+        mask: /^((25[0-5]|2[0-4]\d|[01]?\d?\d)\.){0,3}(25[0-5]|2[0-4]\d|[01]?\d?\d)?$/,
+      }),
+      []
+    );
+
+    const macOptions = useMemo(
+      () => ({
+        mask: Array.from({ length: 6 }).flatMap((_, index) =>
```
lskramarov: **Three near-identical mask builders** _(simplification)_

The MAC, IPv6 and license-key masks are the same expression pasted three times, differing only in group count, group size and separator — and each copy repeats its character regex inline (the license-key one spells `alphanumeric` eight times across three lines).

One local helper covers all three and reads much better in the Source panel, which is what people actually copy:

```tsx
const groups = (count: number, size: number, char: RegExp, separator: string) =>
  Array.from({ length: count }).flatMap((_, index) => [
    ...(index === 0 ? [] : [separator]),
    ...Array.from<RegExp>({ length: size }).fill(char),
  ]);

// mask: groups(6, 2, hex, ':')
// mask: groups(8, 4, hex, ':')
// mask: groups(4, 4, alphanumeric, '-')
```

packages/components/src/components/Input/Input.stories.tsx
```diff
@@ -451,3 +454,148 @@ export const Validation: Story = {
...
+          index === 0 ? [hex, hex, hex, hex] : [':', hex, hex, hex, hex]
+        ),
+      }),
+      []
+    );
+
+    const portOptions = useMemo(
+      () =>
+        maskitoNumberOptionsGenerator({
+          min: 0,
```
lskramarov: **Port bounds `0` / `65535` duplicated in three places** _(simplification)_

The range is written out three times: here in `maskitoNumberOptionsGenerator`, again in `Math.min(Math.max(…, 0), 65535)` inside `handlePortKeyDown`, and a third time in the caption string.

Changing the range means editing three unlinked places; miss one and the mask, the stepper and the visible caption disagree — e.g. with a lowered `max` the keyboard stepper would let the value exceed it until blur, when the min/max plugin clamps it back.

A single `const MIN_PORT = 0;` / `const MAX_PORT = 65535;` that all three derive from (caption via a template literal) removes the drift.

packages/components/src/components/Input/Input.stories.tsx
```diff
@@ -451,3 +454,148 @@ export const Validation: Story = {
...
+      const step = event.key === 'ArrowUp' ? 1 : -1;
+
+      setPort((prevPort) =>
+        String(Math.min(Math.max(Number(prevPort || 0) + step, 0), 65535))
+      );
+    };
+
+    return (
+      <FlexBox direction="column" gap="l" style={{ inlineSize: 320 }}>
+        <Input
```
lskramarov, line 521: **Masked fields are missing `inputMode` / `autoComplete` / `spellCheck`** _(usability)_

None of the five inputs set these, and they're the standard companions to a mask — so readers copying this example inherit the gap:

- On a phone the Port and IP fields open the full alphabetic keyboard instead of a numeric one (`inputMode="numeric"` / `"decimal"`).
- The MAC and license-key fields fight autocapitalize/autocorrect while the postprocessor forces uppercase.
- Browser autofill and spellcheck can inject values that the mask then silently rewrites (see the IPv4 / IPv6 comments).

Adding `inputMode`, `autoComplete="off"` and `spellCheck={false}` to the masked fields makes the example correct on mobile and safe from autofill.

packages/components/src/components/Input/Input.mdx
```diff
@@ -116,6 +117,24 @@ See the example of how to present the component with a password input.
...
+
+The example uses the [Maskito](https://maskito.dev/) library:
+
+```bash
+pnpm install @maskito/core @maskito/react @maskito/kit
```
lskramarov: **`pnpm install <pkg>` vs. the repo's `pnpm add`** _(conventions)_

Every other doc page in this repo uses `pnpm add` when adding packages — [docs/0-welcome.mdx:27](https://github.com/koobiq/react-components/blob/main/docs/0-welcome.mdx#L27) (`pnpm add @koobiq/design-tokens @koobiq/react-components`) and [docs/5-fonts.mdx:29](https://github.com/koobiq/react-components/blob/main/docs/5-fonts.mdx#L29) (`pnpm add @fontsource/inter @fontsource/jetbrains-mono`).

The command here works, but it's the only page using `pnpm install` with package names, and `pnpm install` is the project-wide install command everywhere else in the repo.

```suggestion
pnpm add @maskito/core @maskito/react @maskito/kit
```

packages/components/src/components/Input/Input.stories.tsx
KamilEmeleev: Fixed. The fixed eight-group array mask is replaced with a regex mask:

```ts
mask: /^([\dA-Fa-f]{0,4}:){0,7}[\dA-Fa-f]{0,4}$/
```

I re-ran your table through `maskitoTransform` (`@maskito/core@5.4.0`) — all four inputs now round-trip unchanged: `2001:db8:85a3:0:0:8a2e:370:7334`, `2001:db8::1`, `::1`, `0:0:0:0:0:0:0:1`. Typing the compressed form works too: `:` is no longer swallowed after a group shorter than four digits. The story default is now the canonical compressed address `2001:db8:85a3::8a2e:370:7334`.

Trade-off: the mask no longer auto-inserts `:` — that is the price of accepting variable-length groups and `::`.

packages/components/src/components/Input/Input.stories.tsx
KamilEmeleev: Partly addressed — the regex mask stays, here is why.

An array mask `[d, d, d, '.', …]` auto-inserts the dots only when every octet is three digits: with a fixed slot per digit a short octet can never be terminated, so `192.168.0.1` becomes untypable and only `192.168.000.001` can be entered. That reads worse than the current behaviour.

Per-octet range checking already comes from the regex — `999` yields `99`, and the third digit of `256` is rejected.

The paste case is inherent to a regex mask (`guessValidValueByRegExp` can only drop characters). Rather than implying otherwise, the docs section now states it: a mask drops what does not fit as the user **types or pastes**, and it is not a validation, with a pointer to the Validation section. If you would rather the example carried it, I can add a preprocessor that rejects an invalid paste outright.

packages/components/src/components/Input/Input.stories.tsx
KamilEmeleev: Fixed — the hand-written stepper is gone. The Port field is now `InputNumber`, which handles read-only itself.

packages/components/src/components/Input/Input.stories.tsx
KamilEmeleev: Agreed. The Port field is now:

```tsx
<InputNumber
  label="Port"
  caption="Range: 0-65535"
  defaultValue={8080}
  minValue={0}
  maxValue={65535}
  formatOptions={{ useGrouping: false }}
  fullWidth
/>
```

So the example demonstrates what the Alert above it recommends instead of contradicting it.

packages/components/src/components/Input/Input.stories.tsx
KamilEmeleev: Fixed by the same change. `InputNumber` provides the semantics out of the box — verified in the DOM: `role="group"` with `aria-roledescription="Number field"` on the wrapper, labelled "Increase Port" / "Decrease Port" buttons, plus `inputmode="numeric"`, `autocomplete="off"` and `autocorrect="off"` on the input.

packages/components/src/components/Input/Input.mdx
KamilEmeleev: Rewritten:

> A mask constrains what can be typed into a field: characters that don't fit the format are dropped as the user types or pastes. It is not a validation — an incomplete value such as `192.168` still passes the mask, so pair it with [Validation](#validation) when the field must be complete.

The anchor resolves to the Validation section on the docs page.

pnpm-lock.yaml
KamilEmeleev: Fixed. `pnpm-lock.yaml` was restored from `origin/main` and regenerated, so the diff is now 36 added lines for `@maskito/core`, `@maskito/kit` and `@maskito/react` with zero deletions. The `eslint-plugin-import` duplicate and the `@vitest/coverage-v8` peer drift are gone.

packages/components/src/components/Input/Input.stories.tsx
KamilEmeleev: Done:

```tsx
const maskGroups = (
  count: number,
  size: number,
  character: RegExp,
  separator: string
) =>
  Array.from({ length: count }).flatMap((_, index) => [
    ...(index === 0 ? [] : [separator]),
    ...Array.from<RegExp>({ length: size }).fill(character),
  ]);
```

Used by the MAC (`maskGroups(6, 2, hex, ':')`) and license-key (`maskGroups(4, 4, alphanumeric, '-')`) masks. IPv6 no longer needs it — it moved to a regex mask, see the other thread.

packages/components/src/components/Input/Input.stories.tsx
KamilEmeleev: Resolved by the `InputNumber` switch: the range now lives in `minValue` / `maxValue` only, with the caption as the single piece of prose next to it. The mask options and the hand-written clamp are gone.

packages/components/src/components/Input/Input.stories.tsx
KamilEmeleev: Added: `inputMode="decimal"` on the IP field, plus `autoComplete="off"` and `spellCheck="false"` on every masked field — checked in the DOM that all three reach the native input. The Port field needs nothing: `InputNumber` already sets `inputmode="numeric"`, `autocomplete="off"` and `autocorrect="off"`.

packages/components/src/components/Input/Input.mdx
KamilEmeleev: Fixed — switched to `pnpm add`.

lskramarov (approved):

PR #477, opened 2026-09-09
refactor(components): attach compound slots with Object.assign (DS-5398)

KamilEmeleev:
Compound components attached slots with a local `CompoundedComponent` type plus a cast:

```tsx
type CompoundedComponent = typeof GridComponent & { Item: typeof GridItem };
export const Grid = GridComponent as CompoundedComponent;
Grid.Item = GridItem;
```

Now, like `TopBar`:

```tsx
export const Grid = Object.assign(GridComponent, {
  Item: GridItem,
});
```

Why: the cast claimed a slot the value only got on the next line, and
`CompoundedComponent` was never exported — so 26 API reports showed
`export const X: CompoundedComponent;` with an `ae-forgotten-export` warning
instead of the actual slots. The reports now list the slots.

Also:

- `TreeComponent` is no longer exported — it was public by accident; use `Tree`.
- `GridProps` and `UsernameProps` are derived from the compound instead of the private root.
- `DropdownMenu` renamed its render function and hoisted its cast, like `Menu` / `Table`.
- `@koobiq/react-core` exports `Pressable` with a nameable `PressableProps`, so
  `Menu.Control` and `DropdownMenu.Pressable` need no assertion (TS4023).
- Slot names are no longer checked against a separate type: a typo fails where
  the slot is used and shows up in the API report.
- AGENTS.md updated, API reports regenerated.

Runtime behavior unchanged: `Object.assign` mutates the same object as before.

## Summary by CodeRabbit

* **New Features**
  * Added the public `Pressable` component and its props type for consistent press interactions and ref forwarding.

* **Refactor**
  * Standardized compound component composition across the component library.
  * Improved generated API declarations with explicit types for compound components and their subcomponents.
  * Updated prop types to derive from the complete exported components, improving TypeScript support and API documentation.


packages/components/src/components/SelectNext/Select.tsx
```diff
@@ -481,24 +479,15 @@ function SelectRender<T extends object, M extends SelectionMode = 'single'>(
...
  */
-export const SelectNext = SelectComponent as CompoundedComponent;
-
-SelectNext.Item = SelectOption;
-SelectNext.Section = SelectSection;
-SelectNext.Divider = Divider;
-SelectNext.ItemText = List.ItemText;
-SelectNext.ItemAddon = List.ItemAddon;
-SelectNext.Tag = Tag;
+export const SelectNext = Object.assign(SelectComponent, {
```
lskramarov, line 486: **`SelectNext.api.md` was not regenerated - and nothing will catch it.**

This file lost its `CompoundedComponent` type, but `tools/public_api_guard/components/SelectNext.api.md:42` still reads:

```ts
// Warning: (ae-forgotten-export) The symbol "CompoundedComponent" needs to be exported by the entry point index.d.ts
//
// @public
export const SelectNext: CompoundedComponent;
```

It documents a symbol that no longer exists anywhere in the tree - the only report still mentioning `CompoundedComponent`.

`check-api` won't fail on it, which is the deeper problem: `SelectNext` is missing from the `components` array in `tools/api-extractor/config.json` (it goes `Select` -> `Sidebar` -> `SidePanel`), and `tools/api-extractor/api-extractor.ts` only iterates that array. So `pnpm approve-api` skipped it and `pnpm check-api` never validates it - the public surface of the non-deprecated replacement for `Select` is unguarded, and its checked-in report has been stale since `chore: added api-guard (#365)`.

Suggest adding `"SelectNext"` to `config.json` and running `pnpm build && pnpm approve-api SelectNext` here.

packages/components/src/components/Tree/Tree.tsx
```diff
@@ -34,22 +34,14 @@ export function TreeComponent<T extends object>({
...
  * with support for keyboard navigation and selection.
  */
-export const Tree = TreeComponent as CompoundedComponent;
-
-TreeComponent.Item = TreeItem;
-TreeComponent.ItemContent = TreeItemContent;
-TreeComponent.ItemContentText = ListItemText;
-TreeComponent.ItemContentAddon = ListItemAddon;
-TreeComponent.LoadMoreItem = TreeLoadMoreItem;
+export const Tree = Object.assign(TreeComponent, {
```
lskramarov, line 43: **Breaking type change: `TreeComponent` loses slots it still has at runtime.**

`TreeComponent` is publicly exported (`export function TreeComponent<T extends object>` on line 19, re-exported through `Tree/index.ts` -> `components/index.ts`), and the old code attached the slots to it *directly*:

```ts
TreeComponent.Item = TreeItem;
TreeComponent.ItemContent = TreeItemContent;
// ...
```

TypeScript's expando inference put those into `typeof TreeComponent`, so they were part of the published API - `Tree.api.md` declared them under `export namespace TreeComponent`. This PR deletes that whole block from the report; only `displayName` survives.

`Object.assign` still mutates the same object, so `TreeComponent.Item` **exists at runtime but no longer type-checks**. Anyone on `TreeComponent.Item` against 0.37.0 breaks on upgrade. That is a breaking API change in a `refactor` PR whose body says "Runtime behavior unchanged" and describes this as `Tree` merely "no longer merg[ing] its slots into the `TreeComponent` namespace" - at the type level, yes; at runtime it still does.

Either keep the namespace declared, or flag the removal explicitly in the PR/changelog.

packages/components/src/components/Popover/Popover.tsx
```diff
@@ -42,18 +42,12 @@ const PopoverComponent = forwardRef<ComponentRef<'div'>, PopoverProps>(
...
 /**
  * Popover component as an overlay whose position is anchored to an element in
  * the user interface.
  */
-export const Popover = PopoverComponent as CompoundedComponent;
-
-Popover.Header = Dialog.Header;
-Popover.Body = Dialog.Body;
-Popover.Footer = Dialog.Footer;
+export const Popover = Object.assign(PopoverComponent, {
```
lskramarov, line 49: **Declaration output now inlines the whole props object for every compound whose root isn't nameable.**

Because `PopoverComponent` is a private const and there is no explicit compound type, TS can't name the inferred `Object.assign` result and expands it structurally. `Popover.api.md` went from

```ts
export const Popover: CompoundedComponent;
```

to ~40 lines inlining every prop (`isOpen`, `defaultOpen`, `control`, `slotProps`, ...) instead of referencing `PopoverProps`. Same for Modal (22 lines), SidePanel (22), ContentPanel (21), EmptyState (15), Form (11), and FormField/Grid/Resizable which now inline `Omit<ForwardRefExoticComponent<Merge<DetailedHTMLProps<HTMLAttributes<HTMLDivElement>, ...>>>, never> & PolymorphicWithRef<...>`.

Cost: a bigger `.d.ts` payload for consumers, noisier hover/IntelliSense, and a public surface pinned structurally - any future prop tweak now produces a large report diff instead of a one-line one.

Note which components stayed compact: Accordion, Menu, Table, TagList, TimeRange, DropdownMenu - the ones whose root type is nameable. Declaring and exporting a named compound type (or exporting the root) for the rest gets that back. Ironically `DropdownMenu` already had exactly that (`export const DropdownMenu: CompoundedComponent = Object.assign(...)`) and this PR removes the annotation.

packages/components/src/components/Grid/types.ts
```diff
@@ -49,3 +56,9 @@ export type GridBaseProps = {
   /** Inline styles. */
   style?: CSSProperties;
 };
+
+export type GridProps<As extends ElementType = 'div'> = AsProps<
```
lskramarov: **This contradicts the rule the same PR adds to AGENTS.md.**

`AGENTS.md:249` (added in this PR):

> Derive the props type from the compounded component (`ComponentPropsWithRef<typeof Component>`), not from the private root, so API Extractor doesn't report the root as a forgotten export.

But this line derives `GridProps` from `AsProps<As, GridBaseProps, ComponentPropsWithRef<As>>`, and `Username/types.ts:88` derives `UsernameProps` from `UsernameBaseProps`. Only `FormField.tsx:68` follows the new rule. Three idioms for one concept land in a single PR, leaving no pattern to copy.

The hand-written `AsProps<...>` form also stops tracking the component: if `polymorphicForwardRef`'s `OnlyAs` constraint on `Grid` is ever narrowed (as `FormField.Input` already does with `'input' | 'textarea'`), `GridProps` keeps advertising the wider `ElementType` and silently diverges.

AGENTS.md
```diff
@@ -241,14 +241,13 @@ const ComponentSlot = forwardRef<HTMLSpanElement, ComponentSlotProps>(
...
-};
-
-export const Component = ComponentRoot as CompoundedComponent;
-Component.Slot = ComponentSlot;
+export const Component = Object.assign(ComponentRoot, {
+  Slot: ComponentSlot,
+});
 ```
 
+`Object.assign` builds the value first, so the compound type is inferred structurally — no local intersection type, and no cast claiming slots the value does not have yet. Derive the props type from the compounded component (`ComponentPropsWithRef<typeof Component>`), not from the private root, so API Extractor doesn't report the root as a forgotten export.
```
lskramarov: **The example above can't satisfy the rule this sentence states.**

The example declares `export type ComponentProps = ComponentPropsWithRef<'div'>;` and feeds it into `forwardRef<HTMLDivElement, ComponentProps>`. Applying the rule to that same example:

```ts
const ComponentRoot = forwardRef<HTMLDivElement, ComponentProps>(...);
export const Component = Object.assign(ComponentRoot, { Slot: ComponentSlot });
export type ComponentProps = ComponentPropsWithRef<typeof Component>; // TS2456
```

...is a circular type reference (`Type alias 'ComponentProps' circularly references itself`). A contributor following the doc literally gets a compile error.

The example needs to show the shape the rule assumes - a separate `ComponentBaseProps` fed to the root, with `ComponentProps` derived from the compound - or the prose needs to say when it applies.

tools/public_api_guard/components/Menu.api.md
```diff
@@ -107,6 +109,12 @@ export type MenuSectionProps<T> = SectionProps<T>;
 
 // Warnings were encountered during analysis:
 //
+// packages/components/dist/components/Menu/Menu.d.ts:7:5 - (ae-forgotten-export) The symbol "Item" needs to be exported by the entry point index.d.ts
```
lskramarov: **The `ae-forgotten-export` warnings were relocated, not removed.**

Across all reports in this PR: **58 warnings removed, 57 added**. Menu gains six here (`Item`, `Section`, `Header`, `Divider`, `ListItemTextProps`, `ListItemAddonProps`); Table gains five (`TableHeader`, `TableBody`, `Column`, `Row`, `Cell`); Autocomplete four; Popover three; Tree two; TagList one.

The reports do now show the slot *names*, which is a real gain. But the stated motivation - getting rid of the `CompoundedComponent` forgotten-export warning - nets out to zero, because the actual root cause is untouched: slot components and their props types are not exported from each component's entry-point index.

The deeper fix is to export those symbols (or at least the props types) from the component `index.ts` files, so API Extractor can name them and the warning block empties out for real.

tools/public_api_guard/components/CheckboxGroup.api.md
```diff
@@ -8,12 +8,16 @@ import { AriaCheckboxGroupProps } from 'react-aria';
...
+import { DataAttributeProps } from '@koobiq/react-core';
+import { DetailedHTMLProps } from 'react';
 import type { DOMAttributes } from '@koobiq/react-core';
 import type { ElementType } from 'react';
 import { ExtendableComponentPropsWithRef } from '@koobiq/react-core';
 import { ForwardRefExoticComponent } from 'react';
+import { HTMLAttributes } from 'react';
+import { Merge } from '@koobiq/react-core';
 import { PolyForwardComponent } from '@koobiq/react-core';
+import { PolymorphicWithRef } from '@koobiq/react-core';
```
lskramarov, line 20: **Phantom imports: eight untouched components got their guarded API rewritten with lines that reference nothing.**

`PolymorphicWithRef` appears here at line 20 and **nowhere else in this file** - same for `DetailedHTMLProps` (line 12), plus `HTMLAttributes` and `Merge`, and `DataAttributeProps` flipped from `import type` to `import`. Identical churn in `DateInput`, `DatePicker`, `Input`, `InputNumber`, `RadioGroup`, `SearchInput` and `TimePicker`, none of whose sources changed.

This is fallout from the `FormField` declaration expansion (see the Popover comment) and goes away if the compound types are named explicitly. Until then, unrelated components' guarded API files churn on every `FormField` change, which erodes the guard's value as a review signal.

packages/components/src/components/Username/types.ts
```diff
@@ -84,3 +84,5 @@ export type UsernameSecondaryHintProps = Omit<
   /** Hint content appended after the secondary line. Usually the site. */
   children?: ReactNode;
 };
+
+export type UsernameProps = UsernameBaseProps;
```
lskramarov: **Redundant alias that no longer tracks the component.**

`UsernameProps` and `UsernameBaseProps` are now two exported names for the identical type, with nothing telling consumers which is canonical.

Previously `UsernameProps = ComponentPropsWithRef<typeof UsernameComponent>`, so it followed the component's ref element automatically. Now, if `UsernameComponent` ever changes from `forwardRef<ComponentRef<'span'>, ...>` to another element, `UsernameProps` keeps advertising `ref?: Ref<HTMLSpanElement>` (inherited from the `ComponentPropsWithRef<'span'>` inside `UsernameBaseProps`) and silently lies.

Either derive it from the compound component as `AGENTS.md:249` now instructs, or drop the alias.

packages/components/src/components/Modal/Modal.tsx
```diff
@@ -145,18 +145,12 @@ const ModalComponent = forwardRef<ModalRef, ModalProps>((props, ref) => {
...
  * Modal is a window that appears in front of app content to provide critical
  * information or ask for a decision.
  */
-export const Modal = ModalComponent as CompoundedComponent;
-
-Modal.Header = Dialog.Header;
-Modal.Body = Dialog.Body;
-Modal.Footer = Dialog.Footer;
+export const Modal = Object.assign(ModalComponent, {
+  Header: Dialog.Header,
```
lskramarov: **Reaching through the `Dialog` compound instead of importing the slot directly.**

`Dialog.tsx` and `ContentPanel.tsx` both import `DialogHeader` / `DialogBody` / `DialogFooter` directly; `Modal`, `Popover` and `SidePanel` go through `Dialog.Header` etc.

The move into the `Object.assign` argument also flips the evaluation order: the `Dialog.*` reads now happen *before* this module's own `export const Modal` binding is initialized, where previously `Modal` was bound first and the slots attached after. There is no import cycle through `Dialog` today (`Dialog.tsx` imports nothing from Modal/Popover/SidePanel), so nothing breaks now - but if one ever appears, this turns a benign "slot not attached yet" into a TDZ `ReferenceError`.


Importing `DialogHeader` / `DialogBody` / `DialogFooter` from `'../Dialog'` and using them directly here (and in `Popover.tsx` / `SidePanel.tsx`) removes both the indirection and the ordering dependency, matching `ContentPanel`.

packages/components/src/components/FormField/FormField.tsx
```diff
@@ -53,27 +53,17 @@ export const FormFieldComponent = polymorphicForwardRef<
...
-type CompoundedComponent = typeof FormFieldComponent & {
-  Label: typeof FormFieldLabel;
-  Error: typeof FormFieldError;
-  Caption: typeof FormFieldCaption;
-  ControlGroup: typeof FormFieldControlGroup;
-  Input: typeof FormFieldInput;
-  InputDate: typeof FormFieldInputDate;
-  Select: typeof FormFieldSelect;
-};
+FormFieldComponent.displayName = 'FormField';
```
lskramarov, line 58: **This file is missing `'use client'`.**

`AGENTS.md`: "Every component `.tsx` starts with `'use client'` (Next.js RSC)." `FormField.tsx:1` starts with `import type { ComponentPropsWithRef, ElementType } from 'react';`.

It matters here: `FormField` calls `useForm()` at line 31, which is a `useContext` consumer - in a Next.js App Router server tree it throws rather than rendering.

Pre-existing rather than introduced here, and the fix is one line while the file is already open. `Form/Form.tsx` (also touched by this PR) has the same gap, along with 13 other component files - worth a follow-up sweep.

packages/components/src/components/Menu/Menu.tsx
```diff
@@ -95,23 +95,15 @@ function MenuRender<T>(props: Omit<MenuProps<T>, 'ref'>, ref: Ref<MenuRef>) {
...
+export const Menu = Object.assign(MenuComponent, {
+  Item,
+  Section,
+  Header,
+  Divider,
+  ItemText: ListItemText,
+  ItemAddon: ListItemAddon,
+  // The assertion keeps `Pressable` nameable in the declaration output,
+  // React Aria keeps its props type internal (TS4023).
+  Control: Pressable as typeof Pressable,
```
lskramarov: **The self-assertion treats the symptom, not the cause.**

`Pressable as typeof Pressable` is a no-op at the type level whose only effect is on declaration emit - the same trick appears as `Pressable: Pressable as typeof Pressable` in `DropdownMenu.tsx`. The comment explains *what* it does, but the underlying issue is that `Pressable` has no nameable public props type.

Worth noting: `Pressable` does not appear anywhere in `tools/public_api_guard/react-core.api.md`, yet `Menu.api.md:22` and `DropdownMenu.api.md` emit `import { Pressable } from '@koobiq/react-core'` into the shipped declarations.

Per `AGENTS.md` ("if something is missing, re-export it from core/primitives first"), the general fix is to re-export `Pressable` from `@koobiq/react-core` with an explicit props type and delete both assertions - otherwise this gets copy-pasted into every future component exposing a Pressable slot.

packages/components/src/components/Tree/Tree.tsx
lskramarov, line 43: **`TreeComponent` keeps the slots at runtime but loses them from its type.**

On `main` the slots were assigned to the function declaration itself (`TreeComponent.Item = TreeItem`), which made TS widen its namespace — `Tree.api.md` declared:

```ts
export namespace TreeComponent {
    var displayName: string;
    var Item: typeof TreeItem;
    var ItemContent: typeof TreeItemContent;
    var ItemContentText: ...;
    var ItemContentAddon: ...;
    var LoadMoreItem: typeof TreeLoadMoreItem;
}
```

`Object.assign` mutates the same object, so all five properties are still on `TreeComponent` at runtime — but TS does not widen through it, and the regenerated report now keeps only `displayName`.

`TreeComponent` is public (`export * from './Tree'` -> `components/index.ts`), so `TreeComponent.Item` becomes TS2339 for consumers while still working at runtime. That is a type-level breaking change and it is not covered by "Runtime behavior unchanged" in the description — worth either calling it out in the PR body or making `TreeComponent` non-public.

packages/components/src/components/SelectNext/Select.tsx
lskramarov, line 486: **`SelectNext.api.md` was not regenerated and is now stale.**

`tools/public_api_guard/components/SelectNext.api.md` still contains:

```ts
// Warning: (ae-forgotten-export) The symbol "CompoundedComponent" needs to be exported by the entry point index.d.ts
export const SelectNext: CompoundedComponent;
```

`CompoundedComponent` no longer exists anywhere in the tree after this PR, and `SelectNext` is publicly exported (`packages/components/src/components/index.ts:35`).

Root cause: `SelectNext` is missing from the `components` array in `tools/api-extractor/config.json` (the array jumps `Select` -> `Sidebar`), so `pnpm check-api` never validates it and `pnpm approve-api SelectNext` is a no-op. The report file has not changed since `471ca080`.

Adding `"SelectNext"` to the config and re-running `pnpm build && pnpm approve-api SelectNext` fixes both the stale file and the guard gap.

packages/components/src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -18,43 +18,30 @@ import {
...
-    ItemAddon: ListItemAddon,
-    Section: DropdownMenuSection,
-    Header: DropdownMenuHeader,
-    SubmenuTrigger: DropdownMenuSubmenuTrigger,
-    Autocomplete: DropdownMenuAutocomplete,
-    Footer: DropdownMenuFooter,
-    Pressable,
-  }
-);
+export const DropdownMenu = Object.assign(DropdownMenuComponent, {
```
lskramarov, line 33: **Dropping the annotation removes the only compile-time check on the slot set.**

On `main` this was:

```ts
export const DropdownMenu: CompoundedComponent = Object.assign(DropdownMenuComponentRender, { ... });
```

The explicit `CompoundedComponent` annotation meant that dropping `Footer`, misspelling `SubmenuTrigger`, or attaching the wrong component errored right here. With the type inferred from the literal, any of those now compiles cleanly and surfaces only as a diff in the generated `.api.md` — and not even there for `SelectNext`, which is unguarded (see the other comment).

That is the trade being made across all 28 components. Fine as a deliberate choice, but worth naming in the description, because `check-api` is now the only thing between a slot typo and a released API change.

packages/components/src/components/Popover/Popover.tsx
lskramarov, line 49: **The public `.d.ts` now inlines the whole props object instead of naming `PopoverProps`.**

`Popover.api.md` goes from one line to a 35-line expansion of every `PopoverProps` member:

```ts
export const Popover: ForwardRefExoticComponent<    {
isOpen?: boolean;
defaultOpen?: boolean;
... 24 more ...
slotProps?: { dialog?: DialogProps; arrow?: ...; };
} & { open?: boolean; } & DataAttributeProps & RefAttributes<HTMLDivElement>> & { Header: ...; Body: ...; Footer: ... };
```

…even though `PopoverProps` is itself a public exported type. Same for `Form`, `Modal`, `SidePanel`, `ContentPanel`, `EmptyState`, `FormField`, `Grid`, `Resizable`.

Note which reports stayed compact: `Menu`, `List`, `Table`, `Select`, `DropdownMenu`, `TagInput`, `TimeRange` — exactly the ones whose root already has an exported named component type (`MenuComponent & { ... }`).

The `ae-forgotten-export` warning was literally "the symbol CompoundedComponent needs to be **exported**". Exporting a named compound type from each `types.ts` fixes it at the cause and keeps the public type readable; deleting the alias fixes the warning by inlining, at the cost of bigger `.d.ts` output, unreadable consumer hovers/errors, and report churn on every unrelated prop edit.

packages/components/src/components/Menu/Menu.tsx
```diff
@@ -95,23 +95,15 @@ function MenuRender<T>(props: Omit<MenuProps<T>, 'ref'>, ref: Ref<MenuRef>) {
...
-export const Menu = MenuComponent as CompoundedComponent;
-
-Menu.Item = Item;
-Menu.Section = Section;
-Menu.Header = Header;
-Menu.Divider = Divider;
-Menu.ItemText = ListItemText;
-Menu.ItemAddon = ListItemAddon;
-Menu.Control = Pressable;
+export const Menu = Object.assign(MenuComponent, {
```
lskramarov, line 99: **The refactor does not actually reduce api-extractor warnings — it trades them.**

Across the regenerated reports: **61 `ae-forgotten-export` warnings removed, 58 added.**

`Menu.api.md` alone swaps one `CompoundedComponent` warning for six:

```
+// ... Menu.d.ts:7:5 - The symbol "Item" needs to be exported by the entry point index.d.ts
+// ... Menu.d.ts:8:5 - The symbol "Section" ...
+// ... Menu.d.ts:9:5 - The symbol "Header" ...
+// ... Menu.d.ts:10:5 - The symbol "Divider" ...
+// ... Menu.d.ts:11:5 - The symbol "ListItemTextProps" ...
+// ... Menu.d.ts:12:5 - The symbol "ListItemAddonProps" ...
```

`Form` +6, `Modal` +5, `Table` +5, `Select` +4, `Autocomplete` +4, `ContentPanel` +3, `DropdownMenu` +2, `List` +2, `Tree` +2, `TagInput` +1.

Unlike `CompoundedComponent`, these name internal slot symbols that are deliberately not public, so they can never be resolved — they become permanent noise in the trailing warning block of a dozen reports, which is the signal reviewers use to spot a genuinely leaked internal type.

AGENTS.md
```diff
@@ -240,14 +240,13 @@ const ComponentSlot = forwardRef<HTMLSpanElement, ComponentSlotProps>(
...
```
lskramarov: **This paragraph contradicts the example directly above it, and neither prop type this PR moved follows it.**

> Derive the props type from the compounded component (`ComponentPropsWithRef<typeof Component>`), not from the private root

The snippet at lines 232-235 does the opposite:

```tsx
export type ComponentProps = ComponentPropsWithRef<'div'>;
const ComponentRoot = forwardRef<HTMLDivElement, ComponentProps>(...)
```

Applying the rule there is circular — `ComponentProps` is the input to `ComponentRoot`.

It is also unusable with the repo's own file convention (`types.ts  # prop types (exported as public API)`), because `types.ts` importing the component from `Component.tsx` creates a cycle. That is presumably why this PR wrote `GridProps` as `AsProps<As, GridBaseProps, ComponentPropsWithRef<As>>` and `UsernameProps` as a bare alias — i.e. zero of the two prop types it moved follow the new sentence.

Either drop the sentence, or restate it: "when the props type lives in the component file, derive it from the compounded value (see `FormFieldProps`); when it lives in `types.ts`, use `AsProps<...>` (see `ResizableProps`, `GridProps`)".

packages/components/src/components/DropdownMenu/DropdownMenu.tsx
```diff
@@ -18,43 +18,30 @@ import {
...
+  Popover: DropdownMenuPopover,
+  Content: DropdownMenuContent,
+  Item: DropdownMenuItem,
+  ItemText: ListItemText,
```
lskramarov, line 37: **The `as typeof X` workaround is applied to `Pressable` only, not to `ItemText`/`ItemAddon` right next to it.**

I verified with a standalone `tsc --declaration` repro that the assertion is genuinely load-bearing (without it: TS4023) and that an unasserted imported component is emitted fully expanded instead of as `typeof X`. That is exactly what happens to these two lines:

```ts
ItemText: ForwardRefExoticComponent<Omit<ListItemTextProps, "ref"> & RefAttributes<HTMLSpanElement>>;
ItemAddon: ForwardRefExoticComponent<Omit<ListItemAddonProps, "ref"> & RefAttributes<HTMLDivElement>>;
Pressable: typeof Pressable;   // the asserted one, one line
```

…plus two new `ae-forgotten-export` warnings for `ListItemTextProps`/`ListItemAddonProps` in both `DropdownMenu.api.md` and `Menu.api.md`.

```suggestion
  ItemText: ListItemText as typeof ListItemText,
```

Same for `ItemAddon` below and for `Menu.tsx:104-105`. Either assert all of them consistently or drop the assertion everywhere and accept the expansion — the current mix means these reports re-expand whenever `ListItemTextProps` changes, which is the fragility the comment above `Pressable` warns about.

packages/components/src/components/Username/types.ts
lskramarov: **`UsernameProps` is now a no-op alias of an already-public type.**

`UsernameBaseProps` is exported from this same file and re-exported publicly, so the package ships two identical public names with nothing to distinguish them — consumers and the docs page have no way to know which to reach for, and a future divergence has to be maintained by hand in two places.

Either follow the guidance this PR added to `AGENTS.md` and derive it:

```ts
// Username.tsx
export type UsernameProps = ComponentPropsWithRef<typeof Username>;
```

…or drop `UsernameProps` and keep `UsernameBaseProps` as the single public name.

packages/components/src/components/Grid/types.ts
lskramarov: **`GridProps` re-derives what `polymorphicForwardRef` already computes.**

`AsProps<As, GridBaseProps, ComponentPropsWithRef<As>>` is exactly the props expression inside `PolymorphicWithRef` (`packages/core/src/utils/polymorphicForwardRef.ts:18-25`), so it is equivalent to the old `ComponentPropsWithRef<typeof GridComponent<As>>` today — but only while `Grid` stays `polymorphicForwardRef<'div', GridBaseProps>` with the default `OnlyAs = ElementType`.

Constrain `Grid`'s `OnlyAs` (as `FormFieldInput` already does with `"input" | "textarea"`) or change its default element, and this alias keeps accepting the old set with nothing to flag the drift. The derived form tracked the component automatically; `FormFieldProps` in this same PR still uses it.

`ResizableProps` already uses `AsProps` too, so if this is the intended convention for props living in `types.ts`, it is worth saying so in the AGENTS.md paragraph (which currently says the opposite).

packages/components/src/components/FormField/FormField.tsx
lskramarov, line 58: **`FormField.tsx` (and `Form.tsx`) are missing `'use client'`.**

`AGENTS.md` states: *"Every component `.tsx` starts with `'use client'` (Next.js RSC)"*. Of the 30 component files this PR touches, these two are the only ones starting with an `import` instead.

Pre-existing, not introduced here — but the build uses `preserveModules`, so each emitted module keeps or lacks its own directive, and both files need it: `FormField` calls `useForm()` and `Form` creates that context, so importing either from a Next.js server component fails with *"You're importing a component that needs useContext"*.

Cheap to fix while the files are already open.

packages/components/src/components/Modal/Modal.tsx
lskramarov: **`Modal`/`Popover`/`SidePanel` read `Dialog.Header` at module-init; `ContentPanel` imports it directly.**

`ContentPanel.tsx` does the safer thing:

```ts
export const ContentPanel = Object.assign(ContentPanelComponent, {
  Header: DialogHeader,
  Body: DialogBody,
  Footer: DialogFooter,
});
```

The three others depend on `Dialog`'s module body having already run. If a cycle is ever introduced between `Dialog` and one of them, `Dialog` is `undefined` here and it fails as a `TypeError` at import time rather than a type error. The direct import also gives bundlers a static binding.

Pre-existing, but these exact lines are being rewritten anyway — switching the three to `DialogHeader`/`DialogBody`/`DialogFooter` would make all four Dialog compounds consistent.

packages/components/src/components/SelectNext/Select.tsx
KamilEmeleev: Good catch on the stale report.

Adding `"SelectNext"` to `config.json` doesn't work — API Extractor crashes:

```
API Extractor: SelectNext — crashed: Internal Error: Unable to follow symbol for "T_1"
```

I reproduced it on the base commit with the pre-PR `Select.tsx`, so the crash is not caused by this refactor (possibly the version gap: the extractor bundles TS 5.9.3, the project is on 6.0.3).

So I deleted `SelectNext.api.md` instead — it documented `CompoundedComponent`, which no longer exists anywhere in the tree, and nothing validates it. Fixing the crash and putting `SelectNext` under the guard needs its own issue.

packages/components/src/components/Tree/Tree.tsx
KamilEmeleev: Agreed on the analysis. `TreeComponent` was public by accident — the slots were attached to it directly, so TypeScript merged them into `namespace TreeComponent` and they showed up twice in the report.

Rather than restoring the namespace, I made the root private, like every other compound root: `TreeComponent` is no longer exported, slots are reached through `Tree`. Called out in the PR description.

packages/components/src/components/Popover/Popover.tsx
KamilEmeleev: Accurate, and leaving it as is on purpose: we don't shape component types to satisfy `check-api`.

I did try the fix. A named root type keeps the report compact, but only if the alias is a plain function type — I tested `ForwardRefExoticComponent<PropsWithoutRef<PopoverProps> & RefAttributes<…>>` both as an annotation and as a cast, and the emit still expands the props object. So for `Popover`/`Modal`/`SidePanel`/`ContentPanel` it would mean hand-written function types replacing `ForwardRefExoticComponent`, plus splitting each root in two (`XRender` for `displayName`, `XComponent` for the cast) — a public type change made for the shape of a report. Not worth it; reverted.

`Menu` and `Table` keep their casts for a different reason: `forwardRef` erases the type parameter and the cast is what brings it back. `DropdownMenu` gets a cast for a third reason — its render function is private, so without a named type `meta` in the stories fails with TS4023.

packages/components/src/components/Grid/types.ts
KamilEmeleev: Fixed — one idiom now. `GridProps` is back to `ComponentPropsWithRef<typeof Grid<As>>` in `Grid.tsx` and `UsernameProps` to `ComponentPropsWithRef<typeof Username>`, so both track the component (including a narrowed `OnlyAs`) instead of a hand-written `AsProps<…>`. The `AsProps` form and the `UsernameBaseProps` alias are gone.

AGENTS.md
KamilEmeleev: Right, the example couldn't satisfy its own rule. Rewritten: the root takes `WidgetBaseProps`, and `WidgetProps` is derived from the compound — no circular reference. The prose now also says that a generic component keeps its `as XComponent` cast, since `forwardRef` erases the type parameter.

tools/public_api_guard/components/Menu.api.md
KamilEmeleev: Confirmed, and the PR description was wrong about it — rewritten.

Current numbers against the base: **293 → 290**. 28 gone (`CompoundedComponent` ×26, plus `GridComponent` and `UsernameComponent`), 25 added for the slot symbols themselves. A wash, as you said.

Two things worth separating, though.

The gain here isn't the warning count — it's that the guard shows the slots at all. `export const Menu: CompoundedComponent;` produced no diff when a slot was added, removed or renamed; now it does.

And the "deeper fix" is mostly not about slots. Of the 290 warnings, **208 are an artifact of the entry point**: each report is generated from that component's own `index.d.ts`, so a symbol that *is* public at the package root still counts as forgotten. `Checkbox.api.md` warns about `FormFieldLabelProps` (15 occurrences across reports), `IconButtonProps` (13), `PopoverProps` (8) — all exported from the root barrel and importable by consumers today. Only ~50 are genuine holes: `Collections` (24: `Item`, `Section`, `Header`, `Divider`, `Cell`, `Column` + their props), `Dialog` (16), `DropdownFooterProps` (4), `ResponsiveValue` (4), `BackdropProps` (2); another 29 are `*DeprecatedProps` types.

None of it reaches consumers: `dtsRollup` is disabled, so the shipped `.d.ts` is plain `tsc` output with real imports, and `check-api` doesn't fail on warnings. Emptying that block for real is its own task — pointing the guard at the package root entry would remove the 208 without touching the public API, and the remaining ~50 need per-symbol decisions (export, or a public alias like the existing `PopoverBodyProps = DialogBodyProps`).

tools/public_api_guard/components/CheckboxGroup.api.md
KamilEmeleev: Correct, and it stays — same decision as the `Popover` thread: naming the compound types would fix this, but at the cost of changing the public type of ten components to suit the report.

To be concrete about the trade: with named root types this churn drops to a single import line per report, and I had it implemented before reverting. The reverted state is the deliberate one.

packages/components/src/components/Username/types.ts
KamilEmeleev: Dropped the alias. `UsernameProps` is derived from the compounded component again — `ComponentPropsWithRef<typeof Username>` in `Username.tsx` — so it follows the root's ref element instead of freezing `Ref<HTMLSpanElement>` from `UsernameBaseProps`.

packages/components/src/components/Modal/Modal.tsx
KamilEmeleev: Fixed. `Modal`, `Popover` and `SidePanel` now import `DialogHeader` / `DialogBody` / `DialogFooter` from `'../Dialog'` and use them directly, matching `ContentPanel` — no indirection through the compound and no evaluation-order dependency.

packages/components/src/components/FormField/FormField.tsx
KamilEmeleev: Added `'use client'` to `FormField.tsx` and `Form.tsx`.

The sweep is 12 more files: `Badge`, `DateSegment`, `DropdownFooter`, `FormContext`, `IconItem`, `PopoverInner`, `SelectedTagsMultiline`, `SelectedTagsResponsive`, `SkeletonBlock`, `SkeletonTypography`, `TimePicker`, `Typography`. Leaving those for a separate PR so this one stays a refactor.

packages/components/src/components/Menu/Menu.tsx
KamilEmeleev: Done in core, both assertions removed.

`@koobiq/react-core` now mirrors the type and re-exports the component with it:

```ts
export interface PressableProps extends PressProps {
  children: ReactElement<DOMAttributes, string>;
}

export const Pressable: ForwardRefExoticComponent<
  PressableProps & RefAttributes<FocusableElement>
> = AriaPressable;
```

`PressProps` is exported by `@react-aria/interactions`, so the mirrored type references nothing internal. The first attempt — `export type PressableProps = ComponentPropsWithRef<typeof Pressable>` — did not work: the alias expands back to React Aria's internal interface and TS4023 comes right back. The `= AriaPressable` assignment is the guard against drift on upgrades.

Side effect you predicted: `Pressable` and `PressableProps` now actually appear in `react-core.api.md`.

tools/public_api_guard/components/Modal.api.md
```diff
@@ -20,29 +20,49 @@ import type { ReactElement } from 'react';
...
+dialog?: DialogProps;
+backdrop?: BackdropProps;
+modal?: ComponentPropsWithRef<"div">;
+};
+} & {
+open?: boolean;
+} & RefAttributes<HTMLDivElement>> & {
+    Header: ForwardRefExoticComponent<Omit<DialogHeaderProps, "ref"> & RefAttributes<HTMLDivElement>>;
+    Body: ForwardRefExoticComponent<Omit<DialogBodyProps, "ref"> & RefAttributes<HTMLDivElement>>;
+    Footer: ForwardRefExoticComponent<Omit<DialogFooterProps, "ref"> & RefAttributes<HTMLDivElement>>;
```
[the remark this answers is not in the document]
KamilEmeleev: Not doing this here.

Slot components and their props types are deliberately not exported from a component's entry point — AGENTS.md: "Expose slot components only through the root component." They became visible in the report because the slots are now spelled out instead of hiding behind `CompoundedComponent`; the underlying situation is unchanged.

Nothing reaches consumers either: `dtsRollup` is disabled in `api-extractor.json`, so the shipped `.d.ts` is plain `tsc` output with real relative imports, and `check-api` does not fail on these warnings.

For scale: across all reports it is 293 → 290 versus the base, and 208 of the 290 are an artifact of each report being generated from its own component `index.d.ts` — e.g. `FormFieldLabelProps` and `IconButtonProps` are exported from the package root and importable today. Emptying that block properly is a separate task; for the genuinely unreachable ones (`Dialog*`, `Collections`) the repo already has a cheaper pattern than exporting the symbols: public aliases such as `PopoverBodyProps = DialogBodyProps`.

tools/public_api_guard/components/Tree.api.md
```diff
@@ -22,10 +22,14 @@ import type { TreeItemProps as TreeItemProps_2 } from '@koobiq/react-primitives'
...
 
-// Warning: (ae-forgotten-export) The symbol "CompoundedComponent" needs to be exported by the entry point index.d.ts
-//
 // @public
-export const Tree: CompoundedComponent;
+export const Tree: typeof TreeComponent & {
+    Item: typeof TreeItem;
+    ItemContent: typeof TreeItemContent;
+    ItemContentText: ForwardRefExoticComponent<Omit<ListItemTextProps, "ref"> & RefAttributes<HTMLSpanElement>>;
+    ItemContentAddon: ForwardRefExoticComponent<Omit<ListItemAddonProps, "ref"> & RefAttributes<HTMLDivElement>>;
```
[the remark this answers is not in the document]
KamilEmeleev: Same as the other thread: not exporting these here.

`TreeItem`, `TreeItemContent` and `TreeLoadMoreItem` are exposed through `Tree.*` by design — AGENTS.md: "Expose slot components only through the root component." `ListItemTextProps` and `ListItemAddonProps` are a different case: they are already public at the package root, and the warning only exists because the report's entry point is `dist/components/Tree/index.d.ts` rather than the package entry.

Consumers are unaffected — `dtsRollup` is disabled, the shipped types come from `tsc` — and `check-api` passes. Type-check and the full Vitest run are green on this branch.

lskramarov (approved):

packages/components/src/components/Tree/Tree.tsx
KamilEmeleev: Exporting `TreeComponent` was a mistake on our side. The root is private now, like every other compound root — slots are reached through `Tree`. The PR description says so explicitly. See https://github.com/koobiq/react-components/pull/477#discussion_r3970576832

packages/components/src/components/DropdownMenu/DropdownMenu.tsx
KamilEmeleev: No assertions left since 9fe9072d: `@koobiq/react-core` exports `Pressable` with a nameable `PressableProps` (https://github.com/koobiq/react-components/pull/477#discussion_r3970587227). `ItemText` / `ItemAddon` stay expanded in the report on purpose — same call as https://github.com/koobiq/react-components/pull/477#discussion_r3970578280

packages/components/src/components/TreeSelect/TreeSelect.tsx
```diff
@@ -487,17 +487,10 @@ function TreeSelectRender<T extends object, M extends SelectionMode = 'single'>(
...
-
-TreeSelect.Item = Tree.Item;
-TreeSelect.ItemContent = Tree.ItemContent;
-TreeSelect.LoadMoreItem = Tree.LoadMoreItem;
-TreeSelect.Tag = Tag;
+export const TreeSelect = Object.assign(TreeSelectComponent, {
+  Item: Tree.Item,
+  ItemContent: Tree.ItemContent,
+  LoadMoreItem: Tree.LoadMoreItem,
+  Tag,
```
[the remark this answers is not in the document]
KamilEmeleev: Not new: `main` has the same `TagProps` warning, it only moved from `types.d.ts` to `TreeSelect.d.ts`. `TagProps` is internal on purpose; the public name is `TreeSelectTagProps`. Same reasoning as https://github.com/koobiq/react-components/pull/477#discussion_r3970588634

packages/components/src/components/SelectNext/Select.tsx
KamilEmeleev: Fixed by `main`: #475 put `SelectNext` under `check-api`, and API Extractor no longer crashes on it. The report was regenerated in 0112c75f and now lists the slots.

packages/components/src/components/FormField/FormField.tsx
KamilEmeleev: Added to `FormField.tsx` and `Form.tsx` in 9fe9072d. The rest goes to a separate PR: https://github.com/koobiq/react-components/pull/477#discussion_r3970585945

packages/components/src/components/Popover/Popover.tsx
KamilEmeleev: Leaving the expansion as is — see https://github.com/koobiq/react-components/pull/477#discussion_r3970578280

packages/components/src/components/Menu/Menu.tsx
KamilEmeleev: Agreed, the count is a wash, and the description no longer claims otherwise. Details and the follow-up: https://github.com/koobiq/react-components/pull/477#discussion_r3970581815

PR #479, opened 2026-09-09
fix(Tooltip): align styles with design guidelines (DS-5340)

KamilEmeleev:
## Summary by CodeRabbit

- **Style**
  - Updated tooltip text styling for a more compact appearance.
  - Removed enforced minimum tooltip dimensions, allowing more flexible sizing.
  - Improved warning tooltip text contrast for better readability.

### DIFF
--- a/packages/components/src/components/Tooltip/Tooltip.module.css
+++ b/packages/components/src/components/Tooltip/Tooltip.module.css
@@ -7,8 +7,6 @@
   position: absolute;
   box-sizing: border-box;
   max-inline-size: 300px;
-  min-inline-size: 48px;
-  min-block-size: 32px;
   color: var(--tooltip-color);
   z-index: var(--kbq-layer-modal);
   border-radius: var(--kbq-size-s);
@@ -66,7 +64,7 @@
 }
 
 .warning {
-  --tooltip-color: var(--kbq-foreground-white);
+  --tooltip-color: var(--kbq-foreground-contrast);
   --tooltip-bg-color: var(--kbq-background-warning);
 }
 
--- a/packages/components/src/components/Tooltip/Tooltip.module.css
+++ b/packages/components/src/components/Tooltip/Tooltip.module.css
@@ -8,7 +8,7 @@
   box-sizing: border-box;
   max-inline-size: 300px;
   min-inline-size: 48px;
-  min-block-size: 36px;
+  min-block-size: 32px;
   color: var(--tooltip-color);
   z-index: var(--kbq-layer-modal);
   border-radius: var(--kbq-size-s);
--- a/packages/components/src/components/Tooltip/Tooltip.tsx
+++ b/packages/components/src/components/Tooltip/Tooltip.tsx
@@ -142,7 +142,7 @@ export const Tooltip = forwardRef<TooltipRef, TooltipProps>((props, ref) => {
       className: clsx(
         s.base,
         s[variant],
-        utilClasses.typography['text-normal']
+        utilClasses.typography['text-compact']
       ),
       ref: domRef,
     },
--- a/packages/components/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/components/src/components/Tooltip/Tooltip.stories.tsx
@@ -23,7 +23,6 @@ const meta = {
   parameters: {
     layout: 'centered',
   },
-  tags: ['status:updated', 'date:2026-05-15'],
 } satisfies Meta<typeof Tooltip>;
 
 export default meta;


packages/components/src/components/Tooltip/Tooltip.module.css
```diff
@@ -8,7 +8,7 @@
   box-sizing: border-box;
   max-inline-size: 300px;
   min-inline-size: 48px;
-  min-block-size: 36px;
+  min-block-size: 32px;
```
lskramarov: **RTL: the arrow gap and the arrow itself land on the wrong physical side.** *(Actual lines: 27–33 and 101–109 — outside this hunk, so anchoring here.)*

`data-placement` is written from `useOverlayPosition`'s returned `placement`, whose type is `PlacementAxis = 'top' | 'bottom' | 'left' | 'right' | 'center'` — **physical**. `@react-aria/overlays` already resolves direction for us (`translateRTL`: `position.replace('start','right').replace('end','left')`), so by the time the value reaches the DOM it is never logical.

But the rules keyed on it use **logical** properties:

```css
.base[data-placement='left'][data-arrow='true']  { margin-inline-end: var(--tooltip-arrow-size); }
.base[data-placement='right'][data-arrow='true'] { margin-inline-start: var(--tooltip-arrow-size); }
.arrow[data-placement='left']  { inset-inline-end: calc(-1.5 * var(--tooltip-arrow-size)); }
.arrow[data-placement='right'] { inset-inline-start: calc(-1.5 * var(--tooltip-arrow-size)); }
```

Under `<html dir="rtl">` (the overlay is portalled to `body`, so it inherits the document direction) `inline-end` flips to the physical left. Repro: `dir="rtl"` + `<Tooltip hideArrow={false} placement="start">` → React Aria emits `data-placement="right"` → `margin-inline-start` resolves to `margin-right`, so the 8px gap opens on the side facing *away* from the trigger and the tooltip sits flush against it; `inset-inline-start: -12px` resolves to `right: -12px`, putting the arrow on the opposite edge of the tooltip, detached and pointing the wrong way.

The block-axis rules (`top`/`bottom`) are fine — block direction does not flip with `dir`.

Fix: use physical properties for these four rules (`margin-right` / `margin-left` / `right` / `left`) with a `/* stylelint-disable-next-line plugin/use-logical-properties-and-values */`, since the selector itself is physical. Add a comment explaining why — `pnpm lint:fix` auto-converts physical to logical and will silently reintroduce this otherwise.

The codebase already treats RTL as supported here: `useOverlayArrowStyle` takes `direction` and has explicit RTL tests.

packages/components/src/components/Tooltip/Tooltip.tsx
```diff
@@ -142,7 +142,7 @@ export const Tooltip = forwardRef<TooltipRef, TooltipProps>((props, ref) => {
       className: clsx(
         s.base,
         s[variant],
-        utilClasses.typography['text-normal']
+        utilClasses.typography['text-compact']
```
lskramarov, line 145: **`offset={0}` is silently turned into `4`.** *(Actual line: 76 — outside this hunk, so anchoring here.)*

```tsx
const offset = showArrow ? offsetProp : offsetProp || 4;
```

`||` treats `0` as absent, so the no-arrow branch (the default, since `hideArrow` defaults to `true`) rewrites a deliberate `offset={0}` into `4`. A consumer asking for a tooltip flush against its trigger silently gets a 4px gap, with no way to express "zero".

Reproducible from the `Offsets` story: type `0` into the `offset` input and the tooltip stays 4px off the trigger.

```tsx
const offset = showArrow ? offsetProp : (offsetProp ?? 4);
```

`??` only substitutes for `null`/`undefined`, which is what "no offset passed" actually means here.

lskramarov, line 145: **Overflow-reveal tooltips now render the revealed text smaller than the text they reveal.**

Two in-repo components use `Tooltip` purely as an "show me the part that got truncated" affordance, and both render the source text at 14px:

- `FileUpload/components/FileUploadItem/FileUploadItem.tsx` — `FileUploadItemName` wraps the ellipsised filename in a `Tooltip` gated on `isOverflowing`, and `FileUploadList.tsx:26` renders the list at `utilClasses.typography['text-normal']` (14px).
- `Navbar/components/NavbarItem.tsx` — the same `{children}` is rendered in the `Link` at `typography['text-normal-medium']` (14px) and passed to the `Tooltip` to be revealed when the navbar is collapsed.

After this change the tooltip shows that same string at `text-compact` (12px / 16px line-height), so the "full" text is visibly smaller than the truncated original it is standing in for. Since the design guideline is about tooltips-as-hints, it's worth confirming with design that the truncation-reveal case is meant to shrink too — if not, these two call sites need a way to opt back into `text-normal` (see the separate note about there being no override point).

lskramarov, line 145: **The `status:updated` badge for this change will not show — the `date:` tag is stale.** *(Actual location: `Tooltip.stories.tsx:26`, a file this PR doesn't touch, so anchoring here.)*

```ts
tags: ['status:updated', 'date:2026-05-15'],
```

`.storybook/manager.tsx:45` gives `status:updated` a 30-day TTL, so `isStatusBadgeExpired` considers this badge expired since 2026-06-14. AGENTS.md: *"For `new`/`updated`, include a `date:YYYY-MM-DD` tag — Storybook uses it to expire temporary badges."* This PR is a user-visible restyle, which is exactly what the badge is for, so the date should be bumped to this PR's date and `status:updated` kept.

packages/components/src/components/Tooltip/Tooltip.module.css
lskramarov: **`32px` has a token, and it's also an undocumented derivation.**

AGENTS.md (Styling Approach): *"Visual values come from `@koobiq/design-tokens` CSS custom properties (`--kbq-*`). Don't hard-code a value that has a token."* `--kbq-size-3xl: 32px` exists. (The old `36px` had no token, so this is newly avoidable — and `TopBar.module.css:6` already does exactly this with `var(--kbq-size-6xl)`.)

Separately, `32` isn't an arbitrary number: it's `--kbq-typography-text-compact-line-height` (16px) + the two `var(--kbq-size-s)` paddings on `.content` (8px × 2). That coupling is what makes the PR internally consistent, but nothing in the file records it — change `.content`'s padding or the typography again and this literal silently stops matching, exactly as it just did.

```suggestion
  min-block-size: var(--kbq-size-3xl);
```

Consider a short comment above it noting the `text-compact` line-height + `.content` padding derivation.

packages/components/src/components/Tooltip/Tooltip.module.css
```diff
@@ -8,7 +8,7 @@
   box-sizing: border-box;
   max-inline-size: 300px;
   min-inline-size: 48px;
```
lskramarov: **`min-inline-size` wasn't revisited alongside `min-block-size`.**

The block axis was recomputed for the compact scale (16px line-height + 2×8px padding = 32px), but the inline axis kept its 14px-era `48px`. With 12px text and 24px of inline padding, that leaves 24px for content — a short tooltip ("On", "12", "Копировать" truncated, etc.) is padded out to 48px, and since `.content` isn't centered (`.container` is a column flex, `.content` a full-width block) the extra space all lands as trailing whitespace. That asymmetry is more noticeable at 12px than it was at 14px.

Either confirm 48px is still the design minimum, or drop it to match the new scale. Whichever value survives, `--kbq-size-5xl` is 48px if it stays.

packages/components/src/components/Tooltip/Tooltip.tsx
lskramarov, line 145: **There's no supported way for a consumer to keep the old size.**

The typography class is baked in here and then merged via React Aria's `mergeProps`, which *concatenates* `className` rather than replacing it (`mergeProps.mjs:34`: `key === 'className' → clsx(a, b)`). So a consumer's own class lands alongside `.text-compact`, not instead of it, and whichever rule wins comes down to emission order in `dist/style.css` — in practice they need `!important` or a specificity hack.

AGENTS.md's documented escape hatch is a public `--kbq-<component>-*` variable, but `Tooltip.module.css` declares only private ones (`--tooltip-bg-color`, `--tooltip-color`, `--tooltip-arrow-size`) and wires none of them to a `--kbq-tooltip-*` override, so the component currently exposes no override points at all.

`Tooltip` is `<Status variant="experimental" />` and the lifecycle doc allows breaking changes in minor releases, so this isn't a blocker — but since this PR is the change that makes someone want the old value, it's the natural moment to add a `--kbq-tooltip-*` hook (at minimum for font-size/line-height, ideally for the three existing private vars too).

packages/components/src/components/Tooltip/Tooltip.module.css
lskramarov: **Smaller text compounds the already-failing contrast on the `warning` variant.** *(Actual lines: 68–71 — outside this hunk, so anchoring here.)*

```css
.warning {
  --tooltip-color: var(--kbq-foreground-white);
  --tooltip-bg-color: var(--kbq-background-warning);
}
```

In the light theme `--kbq-background-warning` is `oklch(80.8% 0.1990 71.4 / 0.69)` — white on a light orange, roughly 1.7:1, well under the WCAG AA 4.5:1 for body text. The threshold doesn't move (12px and 14px are both "normal text", so 4.5:1 applies either way), but 12px makes an already-hard-to-read combination materially harder.

Pre-existing rather than introduced here, so not a blocker for this PR — but since this is the change that shrinks the text, worth raising with design whether `warning` (and `theme`, at `oklch(50.6% … / 0.73)`) should switch to a dark foreground, or whether those variants should stay at `text-normal`.

KamilEmeleev: Fixed: `warning` now uses `--kbq-foreground-contrast` for text, as in the Angular reference (`--kbq-tooltip-warning-fade-off-text`). `theme` stays white — it matches the reference too.

KamilEmeleev: Removed `min-block-size` altogether: the content already gives 32px (16px line-height + 2 × 8px padding), and the Angular reference has no min-height, so there's no number left to keep in sync.

packages/components/src/components/Tooltip/Tooltip.module.css
KamilEmeleev: Removed `min-inline-size` as well — the Angular reference has no min-width, so short tooltips now fit their text.

packages/components/src/components/Tooltip/Tooltip.tsx
KamilEmeleev: Intended: the design token `--kbq-tooltip-font-default-*` is `text-compact`, and the Angular tooltip uses `text-compact` everywhere, including the navbar. So the truncation-reveal case shrinks too.

KamilEmeleev: Removed the stale `status:updated` / `date:` tags instead of bumping the date.

packages/components/src/components/Tooltip/Tooltip.module.css
KamilEmeleev: Confirmed, thanks. It's pre-existing and `Popover` has the same rules, so both go to a separate task (DS-5557).

packages/components/src/components/Tooltip/Tooltip.tsx
KamilEmeleev: Confirmed, and the JSDoc says `@default 0` as well. Pre-existing, so it goes to a separate fix: `??` plus a JSDoc update (DS-5558).

KamilEmeleev: Agreed it's worth having, but it's new public API, so it goes to a separate task (DS-5559).

lskramarov (approved):

PR #480, opened 2026-09-09
fix(Popover, Modal, SidePanel): unify the overlay z-index layer (DS-5504)

KamilEmeleev:
## Summary by CodeRabbit

* **Bug Fixes**
  * Standardized modals, popovers, tooltips, file upload overlays, and side panels on a shared overlay stacking layer.
  * Ensured popovers honor the shared layer while still allowing explicit custom stacking values.

* **Documentation**
  * Added guidance and an interactive example explaining overlay stacking order, including toast behavior and application stacking limits.

* **Tests**
  * Added coverage validating overlay ordering and stacking behavior.

### DIFF
--- a/packages/components/src/components/Modal/Modal.mdx
+++ b/packages/components/src/components/Modal/Modal.mdx
@@ -105,3 +105,12 @@ You can adjust the interaction with the modal window using the following propert
 - `disableExitOnEscapeKeyDown` — if `true`, the component won't close when the ESC key is pressed.
 
 <Story of={Stories.Settings} />
+
+## Overlay stacking
+
+All overlays share one layer, `--kbq-layer-overlay` (`1000`), so the last one
+opened is on top. Toasts are the exception: `--kbq-layer-toast` is above them all.
+
+Keep your app's own `z-index` values below `1000`.
+
+<Story of={Stories.Stacking} />
--- a/packages/components/src/components/Modal/Modal.mdx
+++ b/packages/components/src/components/Modal/Modal.mdx
@@ -108,9 +108,10 @@ You can adjust the interaction with the modal window using the following propert
 
 ## Overlay stacking
 
-All overlays share one layer, `--kbq-layer-overlay` (`1000`), so the last one
-opened is on top. Toasts are the exception: `--kbq-layer-toast` is above them all.
+All overlays share one layer, `--kbq-layer-overlay`, so the last one opened is
+on top. Toasts are the exception: `--kbq-layer-toast` is above them all.
 
-Keep your app's own `z-index` values below `1000`.
+Keep your app's own `z-index` values below `--kbq-layer-topbar`, the lowest
+layer the library occupies.
 
 <Story of={Stories.Stacking} />
--- a/packages/components/src/components/Modal/Modal.stories.tsx
+++ b/packages/components/src/components/Modal/Modal.stories.tsx
@@ -1,4 +1,4 @@
-import { useBoolean } from '@koobiq/react-core';
+import { useBoolean, useTimer } from '@koobiq/react-core';
 import type { Meta, StoryObj } from '@storybook/react';
 
 import { Button } from '../Button';
@@ -24,7 +24,6 @@ const meta = {
   parameters: {
     layout: 'centered',
   },
-  tags: ['status:updated', 'date:2026-09-09'],
 } satisfies Meta<typeof Modal>;
 
 export default meta;
@@ -218,6 +217,12 @@ export const Stacking: Story = {
   render: function Render() {
     const [isOpen, { on, set }] = useBoolean(false);
 
+    const { count, isTimerRunning, startTimer } = useTimer({
+      startTime: 1500,
+      interval: 500,
+      onTimerEnd: on,
+    });
+
     const items = [
       <Select.Item id="bruteforce" key="bruteforce">
         Bruteforce
@@ -242,8 +247,10 @@ export const Stacking: Story = {
         >
           {items}
         </Select>
-        <Button onPress={() => setTimeout(on, 1500)}>
-          Open the modal in 1.5s
+        <Button onPress={startTimer}>
+          {isTimerRunning
+            ? `Opening the modal in ${count / 1000}s`
+            : 'Open the modal in 1.5s'}
         </Button>
         <Modal isOpen={isOpen} size="small" onOpenChange={set}>
           <Modal.Header>Overlay stacking</Modal.Header>
--- a/packages/components/src/components/Modal/Modal.stories.tsx
+++ b/packages/components/src/components/Modal/Modal.stories.tsx
@@ -5,8 +5,10 @@ import { Button } from '../Button';
 import { FlexBox } from '../FlexBox';
 import { Form } from '../Form';
 import { Input } from '../Input';
+import { SelectNext as Select } from '../SelectNext';
 import { Textarea } from '../Textarea';
 import { Toggle } from '../Toggle';
+import { Typography } from '../Typography';
 
 import { Modal } from './index';
 import { modalPropSize } from './index.js';
@@ -22,6 +24,7 @@ const meta = {
   parameters: {
     layout: 'centered',
   },
+  tags: ['status:updated', 'date:2026-09-09'],
 } satisfies Meta<typeof Modal>;
 
 export default meta;
@@ -207,6 +210,60 @@ export const Settings: Story = {
           </>
         )}
       </Modal>
+    );
+  },
+};
+
+export const Stacking: Story = {
+  render: function Render() {
+    const [isOpen, { on, set }] = useBoolean(false);
+
+    const items = [
+      <Select.Item id="bruteforce" key="bruteforce">
+        Bruteforce
+      </Select.Item>,
+      <Select.Item id="ddos" key="ddos">
+        DDoS
+      </Select.Item>,
+      <Select.Item id="identity-theft" key="identity-theft">
+        Identity Theft
+      </Select.Item>,
+      <Select.Item id="network-attack" key="network-attack">
+        Network Attack
+      </Select.Item>,
+    ];
+
+    return (
+      <FlexBox gap="l" direction="column" alignItems="flex-start">
+        <Select
+          label="Attack type"
+          style={{ inlineSize: 200 }}
+          placeholder="Select an option"
+        >
+          {items}
+        </Select>
+        <Button onPress={() => setTimeout(on, 1500)}>
+          Open the modal in 1.5s
+        </Button>
+        <Modal isOpen={isOpen} size="small" onOpenChange={set}>
+          <Modal.Header>Overlay stacking</Modal.Header>
+          <Modal.Body>
+            <FlexBox gap="l" direction="column" alignItems="flex-start">
+              <Typography variant="text-normal">
+                The dropdown above opened before this modal, so the modal covers
+                it. The one below opens last and stays on top.
+              </Typography>
+              <Select
+                label="Attack type"
+                style={{ inlineSize: 200 }}
+                placeholder="Select an option"
+              >
+                {items}
+              </Select>
+            </FlexBox>
+          </Modal.Body>
+        </Modal>
+      </FlexBox>
     );
   },
 };
--- a/packages/components/src/components/Modal/Modal.test.tsx
+++ b/packages/components/src/components/Modal/Modal.test.tsx
@@ -5,6 +5,7 @@ import { userEvent } from '@testing-library/user-event';
 import { describe, expect, it, vi, beforeEach } from 'vitest';
 
 import { Button } from '../Button';
+import { Popover } from '../Popover';
 
 import { Modal } from './Modal';
 import { type ModalProps, modalPropSize } from './types';
@@ -307,6 +308,52 @@ describe('Modal', () => {
       expect(onOpenChange).toHaveBeenCalledTimes(1);
 
       expect(onOpenChange.mock.results[0]?.value).toStrictEqual(false);
+    });
+  });
+
+  // Every overlay shares one z-index layer, so the DOM order of the portals
+  // decides which one is on top: whatever opens last wins.
+  describe('overlay stacking', () => {
+    it('should be portaled after a popover that was already open', () => {
+      const { rerender } = render(
+        <>
+          <Popover data-testid="popover" isOpen />
+          <Modal {...baseProps} isOpen={false} />
+        </>
+      );
+
+      rerender(
+        <>
+          <Popover data-testid="popover" isOpen />
+          <Modal {...baseProps} isOpen />
+        </>
+      );
+
+      const position = screen
+        .getByTestId('popover')
+        .compareDocumentPosition(getModal());
+
+      expect(position & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
+    });
+
+    it('should be portaled before a popover opened from inside it', () => {
+      const { rerender } = render(
+        <Modal {...baseProps} isOpen>
+          <Popover data-testid="popover" isOpen={false} />
+        </Modal>
+      );
+
+      rerender(
+        <Modal {...baseProps} isOpen>
+          <Popover data-testid="popover" isOpen />
+        </Modal>
+      );
+
+      const position = getModal().compareDocumentPosition(
+        screen.getByTestId('popover')
+      );
+
+      expect(position & Node.DOCUMENT_POSITION_FOLLOWING).toBeTruthy();
     });
   });
 });
--- a/packages/components/src/components/Popover/Popover.module.css
+++ b/packages/components/src/components/Popover/Popover.module.css
@@ -3,6 +3,7 @@
   --popover-border-radius: var(--kbq-size-m);
 
   display: flex;
+  z-index: var(--kbq-layer-overlay);
   border-radius: var(--kbq-popover-border-radius, var(--popover-border-radius));
   box-shadow: var(--kbq-shadow-overlay);
   background-color: var(--kbq-background-bg);
--- a/packages/components/src/components/Popover/PopoverInner.tsx
+++ b/packages/components/src/components/Popover/PopoverInner.tsx
@@ -151,6 +151,14 @@ export const PopoverInner: FC<PopoverInnerProps> = (props) => {
       ? children({ close: state.close })
       : children;
 
+  const popoverStyle = {
+    ...popoverProps.style,
+    // React Aria hardcodes `z-index: 100000` here. Strip it so the layer comes
+    // from CSS and every overlay shares one stacking plane, where the DOM order
+    // of the portals decides. A `style` prop from above still wins.
+    zIndex: props.style?.zIndex,
+  };
+
   const renderPopover = (transition: string) => (
     <div
       ref={domRef}
@@ -163,7 +171,7 @@ export const PopoverInner: FC<PopoverInnerProps> = (props) => {
       style={
         {
           ...props.style,
-          ...popoverProps.style,
+          ...popoverStyle,
           '--popover-inline-size': normalizeInlineSize(size),
         } as CSSProperties
       }
--- a/packages/components/src/components/Popover/PopoverInner.tsx
+++ b/packages/components/src/components/Popover/PopoverInner.tsx
@@ -151,14 +151,6 @@ export const PopoverInner: FC<PopoverInnerProps> = (props) => {
       ? children({ close: state.close })
       : children;
 
-  const popoverStyle = {
-    ...popoverProps.style,
-    // React Aria hardcodes `z-index: 100000` here. Strip it so the layer comes
-    // from CSS and every overlay shares one stacking plane, where the DOM order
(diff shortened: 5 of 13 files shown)


packages/components/src/components/Popover/Popover.module.css
```diff
@@ -3,6 +3,7 @@
   --popover-border-radius: var(--kbq-size-m);
 
   display: flex;
+  z-index: var(--kbq-layer-overlay);
```
lskramarov, line 6: **Popover-based overlays now paint under the toast layer.**

Dropping every popover from React Aria's inline `z-index: 100000` to `--kbq-layer-overlay` (1000) puts them below `ToastRegion` (`ToastRegion.module.css:5`, `--kbq-layer-toast` = 1100), which is `position: fixed` and forms its own stacking context.

`ToastContentProps.action` is documented as *"Action element, such as a button or link"*. A toast whose action opens a `DropdownMenu` / `SelectNext` / `Menu` / `DatePicker` now has that popover portaled to `document.body` at 1000 — painted **underneath the toast that opened it**. Before this PR it was at 100000 and rendered above. The same happens to any dropdown that visually overlaps a top/bottom-centred toast stack.

The docs say "Toasts continue to appear above other overlays", which is intended for modals — but an overlay opened *from inside* a toast looks like an unintended casualty.

packages/components/src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -34,6 +34,12 @@ describe('Tooltip', () => {
     expect(ref.current).toBe(getRoot());
   });
 
+  it('should not inline the react-aria z-index, so the CSS layer applies', () => {
```
lskramarov: **This test can never fail — and it hides a real bug.**

`Tooltip.tsx:128` funnels `overlayProps` through `useTooltip`, which runs `filterDOMProps(props, { labelable: true })`. That keeps only `id`, `aria-label*`/`aria-describedby`/`aria-details` and `data-*` — `style` is dropped wholesale, not just `zIndex`.

I rendered `<Tooltip isOpen />` and dumped the root: the `style` attribute is **`null`**. (The popover root, for comparison, gets `position: fixed; top: 0px; left: 0px; max-height: 100vh`.) So `getRoot().style.zIndex === ''` was already unconditionally true before this PR — `PopoverInner` got a real fix, `Tooltip` got a test asserting a fix it never received.

The bigger problem: the same filtering also discards `position` / `top` / `left` / `maxHeight` from `useOverlayPosition`, so the tooltip is never inline-positioned at its anchor and falls back to its static position under `.base { position: absolute }`.

Suggested fix: apply `overlayProps.style` to the tooltip element directly (minus `zIndex`), the way `PopoverInner` now does — then this test starts guarding something.

packages/components/src/global.css
```diff
@@ -3,9 +3,12 @@
   --kbq-layer-default: 0;
   --kbq-layer-absolute: 1;
   --kbq-layer-topbar: 950;
-  --kbq-layer-modal: 1000;
+  --kbq-layer-overlay: 1000;
   --kbq-layer-toast: 1100;
 
+  /** @deprecated Use `--kbq-layer-overlay` instead. */
+  --kbq-layer-modal: var(--kbq-layer-overlay);
```
lskramarov: **The alias preserves reads but not writes — existing overrides break silently.**

`--kbq-layer-*` are `--kbq-`-prefixed public custom properties (AGENTS.md: *"z-index layers are the `--kbq-layer-*` variables from `global.css`"*). An app with

```css
:root { --kbq-layer-modal: 2000; }
```

used to lift Modal, SidePanel, Tooltip and the FileUpload drop overlay above its own chrome. After this PR every component reads `var(--kbq-layer-overlay)` = 1000, so that override does nothing — no error, no console deprecation, nothing visible until a QA pass.

Inverting the alias keeps both directions working:

```css
/** @deprecated Use `--kbq-layer-overlay` instead. */
--kbq-layer-modal: 1000;

--kbq-layer-overlay: var(--kbq-layer-modal);
```

Then a legacy `--kbq-layer-modal` override still flows through to every overlay, and new code sets `--kbq-layer-overlay`.

packages/components/src/components/Modal/Modal.stories.tsx
```diff
@@ -210,3 +213,57 @@ export const Settings: Story = {
     );
   },
 };
+
+export const Stacking: Story = {
```
lskramarov, line 216: **This story can't demonstrate its stated premise — the outer dropdown closes when the modal opens.**

Verified in jsdom: open the outer `SelectNext`, then open the `Modal` — `queryAllByRole('listbox')` goes **1 → 0**.

`usePopover` passes `shouldCloseOnBlur: true` to `useOverlay`, and `Modal`'s `Overlay` wraps its children in a focus-containing `FocusScope`. Pulling focus into the modal blurs the open listbox, which dismisses it.

So the prose here and at `Modal.mdx:111` — *"The dropdown above opened before this modal, so the modal covers it"* — describes something the docs page never actually shows. Only the second half (the dropdown opened last stays on top) is observable, and the 1.5s `setTimeout` device buys nothing.

Consider dropping the outer Select and demonstrating the ordering some other way, e.g. two stacked overlays the reader opens in sequence and can keep open simultaneously.

packages/components/src/components/Popover/PopoverInner.tsx
```diff
@@ -151,6 +151,14 @@ export const PopoverInner: FC<PopoverInnerProps> = (props) => {
       ? children({ close: state.close })
       : children;
 
+  const popoverStyle = {
```
lskramarov: **Consider stripping the z-index one layer down, in `@koobiq/react-primitives`.**

The root cause is React Aria hardcoding `zIndex: 100000` in `useOverlayPosition` ([`useOverlayPosition.mjs`](https://unpkg.com/@react-aria/overlays@3.31.2/dist/useOverlayPosition.mjs)); this strips it at one component call site.

AGENTS.md designates `@koobiq/react-primitives` as *"the React Aria integration layer"* and forbids components from reaching React Aria directly. A primitives-level wrapper around `usePopover` / `useOverlayPosition` that omits `zIndex` would cover `Tooltip` (which calls `useOverlayPosition` itself) and every future overlay for free.

As written, the next component that positions an overlay silently reintroduces `z-index: 100000` and breaks the "one shared plane, DOM order decides" contract this PR is establishing — and no test would notice.

packages/components/src/components/Modal/Modal.test.tsx
```diff
@@ -309,4 +310,50 @@ describe('Modal', () => {
       expect(onOpenChange.mock.results[0]?.value).toStrictEqual(false);
     });
   });
+
+  // Every overlay shares one z-index layer, so the DOM order of the portals
+  // decides which one is on top: whatever opens last wins.
+  describe('overlay stacking', () => {
+    it('should be portaled after a popover that was already open', () => {
```
lskramarov, line 317: **Both of these tests pass on `main` too — they can't fail for the reason this PR exists.**

An overlay that mounts later is *always* appended later in `document.body`; that's a property of `createPortal` plus `unmountOnExit`, independent of z-index. Set `Modal.module.css` back to a layer above `--kbq-layer-overlay`, or give `Popover` a different one, and both tests stay green while the shipped stacking is broken.

Nothing in the suite asserts what the PR actually changes: that Modal / SidePanel / Tooltip / FileUploadDropTargetOverlay all resolve to `--kbq-layer-overlay`, or that the deprecated `--kbq-layer-modal` alias still resolves to the same value. `Popover.test.tsx:53` is the one new test that does guard the fix.

A cheap addition in this file: assert the modal root has no inline `z-index` and that its class list carries the shared layer class — or, since CSS Modules values aren't resolvable in jsdom, a small snapshot/unit check over `global.css` that `--kbq-layer-modal` and `--kbq-layer-overlay` agree.

packages/components/src/components/Modal/Modal.stories.tsx
```diff
@@ -22,6 +24,7 @@ const meta = {
   parameters: {
     layout: 'centered',
   },
+  tags: ['status:updated', 'date:2026-09-09'],
```
lskramarov: **`Popover` is the component whose behaviour actually changed, but only `Modal` is tagged and documented.**

AGENTS.md: *"Component stories (`*.stories.tsx`) use the `Components` group and `tags` such as `status:new`, `status:updated`, `status:deprecated`. For `new`/`updated`, include a `date:YYYY-MM-DD` tag."*

`Popover.stories.tsx` carries no `tags` at all, and `Popover.mdx` gains nothing — so a consumer on Popover's page sees no badge and no mention that its stacking dropped from 100000 to 1000, nor that `style={{ zIndex }}` is now the supported override (behaviour this PR added a test for at `Popover.test.tsx:59` but never documented). `SidePanel.stories.tsx` is likewise untagged.

Modal's own stacking value didn't change here — only its variable name did.

packages/components/src/components/Modal/Modal.mdx
```diff
@@ -105,3 +105,12 @@ You can adjust the interaction with the modal window using the following propert
 - `disableExitOnEscapeKeyDown` — if `true`, the component won't close when the ESC key is pressed.
 
 <Story of={Stories.Settings} />
+
+## Overlay stacking
+
+All overlays share one layer, `--kbq-layer-overlay` (`1000`), so the last one
+opened is on top. Toasts are the exception: `--kbq-layer-toast` is above them all.
+
+Keep your app's own `z-index` values below `1000`.
```
lskramarov: **This is a migration requirement for a breaking change, shipped as a `fix`.**

Popover-based overlays previously escaped essentially any app stacking context at `z-index: 100000`. Any consumer with the very common `z-index: 9999` sticky header, a third-party chat widget, or a cookie banner will now have Select/DropdownMenu/DatePicker popovers render underneath after a patch upgrade. Per the repo's Git Commit Convention only `feat` and `fix` reach the changelog, and a `fix` entry reads as safe to take.

Two smaller things on this line:

- It hardcodes `1000` in prose, so it goes stale if the token moves — prefer naming the token only.
- It contradicts the shipped `--kbq-layer-topbar: 950`: an app that dutifully keeps its own chrome "below 1000" at, say, 990 now covers the Koobiq TopBar. The real ceiling for app content is 950.

packages/components/src/components/Popover/PopoverInner.tsx
```diff
@@ -151,6 +151,14 @@ export const PopoverInner: FC<PopoverInnerProps> = (props) => {
...
+    ...popoverProps.style,
+    // React Aria hardcodes `z-index: 100000` here. Strip it so the layer comes
+    // from CSS and every overlay shares one stacking plane, where the DOM order
+    // of the portals decides. A `style` prop from above still wins.
+    zIndex: props.style?.zIndex,
```
lskramarov: **The `popoverStyle` intermediate isn't needed — the final literal already spreads `props.style` first.**

Today the reader has to hold two objects in their head to see that `props.style` is spread, then overridden by `popoverProps.style`, then partially re-overridden for `zIndex` only. Collapsing it keeps the whole ordering on one screen:

```tsx
      style={
        {
          ...props.style,
          ...popoverProps.style,
          // React Aria hardcodes `z-index: 100000` here. Strip it so the layer
          // comes from CSS and every overlay shares one stacking plane, where
          // the DOM order of the portals decides. A `style` prop still wins.
          zIndex: props.style?.zIndex,
          '--popover-inline-size': normalizeInlineSize(size),
        } as CSSProperties
      }
```

Same output, one fewer indirection, and the intent reads off the `zIndex` line itself.

packages/components/src/components/Modal/Modal.stories.tsx
```diff
@@ -210,3 +213,57 @@ export const Settings: Story = {
...
+    return (
+      <FlexBox gap="l" direction="column" alignItems="flex-start">
+        <Select
+          label="Attack type"
+          style={{ inlineSize: 200 }}
+          placeholder="Select an option"
+        >
+          {items}
+        </Select>
+        <Button onPress={() => setTimeout(on, 1500)}>
```
lskramarov: **This timer is never cleared.**

Press the button twice and switch stories before 1.5s elapses: both timers still fire and call `set` on the unmounted `Render`. Press it repeatedly while the modal is closing and it re-opens on each stale timer.

If the delay survives the other feedback on this story, hold the id in a ref and clear it in a `useEffect` cleanup.

packages/components/src/components/Modal/Modal.mdx
```diff
@@ -105,3 +105,12 @@ You can adjust the interaction with the modal window using the following propert
 - `disableExitOnEscapeKeyDown` — if `true`, the component won't close when the ESC key is pressed.
 
 <Story of={Stories.Settings} />
+
+## Overlay stacking
```
lskramarov, line 109: **This contract is cross-cutting, but it only lives on Modal's page.**

It governs Popover, Tooltip, SidePanel, FileUpload and Toast. A consumer debugging why their Select dropdown sits under their app header opens `Popover.mdx` and finds nothing about layers, the shared `--kbq-layer-overlay` plane, the "last opened wins" rule, or the `style={{ zIndex }}` escape hatch.

`docs/` already hosts numbered cross-cutting guides (`3-responsive-ui.mdx`, `9-deprecation-strategy.mdx`). A `docs/N-overlay-layers.mdx` covering the layer table, the DOM-order rule, the toast exception, the `portalContainer` caveat (ordering only holds while every overlay portals to the same container) and the `--kbq-layer-modal` deprecation would be the right home, with short cross-links from the affected component pages.

packages/components/src/components/Popover/Popover.module.css
lskramarov, line 6: **`.underlay` (line 77) was left out of the unification.**

It keeps `z-index: auto` while `.base` moves onto the shared 1000 plane. The underlay is a sibling of the popover inside the same portal with only `inset: 0; position: fixed`, and positioned elements with `z-index: auto` paint before any positive z-index.

So with a Modal open (`.base` at 1000) and a modal-type Popover opened on top of it, the popover paints above the modal but its click-blocking underlay paints below it — clicks on the modal body hit the modal, not the popover's underlay. Dismissal still works because React Aria listens at the document level, but the underlay no longer does the job it exists for.

Giving `.underlay` the same `z-index: var(--kbq-layer-overlay)` keeps the pair together on the shared plane.

packages/components/src/global.css
KamilEmeleev: Good catch, fixed. Flipped the alias so the old name stays the override point:

```css
--kbq-layer-modal: 1000; /* @deprecated Use `--kbq-layer-overlay`. */
--kbq-layer-overlay: var(--kbq-layer-modal);
```

Checked both directions in the browser on an open Select popover: no override → 1000, `--kbq-layer-modal: 2000` → 2000, `--kbq-layer-overlay: 3000` → 3000.

packages/components/src/components/Modal/Modal.test.tsx
KamilEmeleev: You're right, those two only prove `createPortal` appends in mount order. Added `src/global.test.ts` to guard the actual wiring: all five overlay stylesheets must reference the same `--kbq-layer-*` token, and `--kbq-layer-overlay` must still resolve through the deprecated alias.

Verified it goes red both ways — moved `Modal.module.css` to another layer and broke the alias, each fails its own test.

Kept the portal-order tests, they still document the "last opened wins" half of the contract.

packages/components/src/components/Popover/PopoverInner.tsx
KamilEmeleev: Agreed, collapsed it into the literal.

packages/components/src/components/Modal/Modal.stories.tsx
KamilEmeleev: Fixed — swapped the raw `setTimeout` for `useTimer` from `@koobiq/react-core`. It cleans up on unmount, restarts on repeated presses, and gives the button a visible countdown so the story explains itself.

packages/components/src/components/Modal/Modal.stories.tsx
KamilEmeleev: Fair point that Modal's value didn't change here. We decided not to tag either component, so I dropped the `status:updated` tag from Modal rather than moving it to Popover.

packages/components/src/components/Modal/Modal.mdx
KamilEmeleev: The 950 contradiction is a good catch — reworded to point at `--kbq-layer-topbar` as the ceiling, and dropped the hardcoded number so it can't go stale:

> Keep your app's own `z-index` values below `--kbq-layer-topbar`, the lowest layer the library occupies.

On `fix` vs breaking: we read this as a bug rather than an API change. A popover was never meant to cover a modal — that was the defect. The layer tokens keep their values, the public props are untouched, and apps that pin their own chrome under the library's layers see no difference.

packages/components/src/components/Tooltip/Tooltip.test.tsx
KamilEmeleev: First half is right — the test was green before this branch too, so it guarded nothing. Removed it.

The positioning part doesn't hold though. `useOverlayPosition` writes the coordinates straight to the DOM node, it doesn't go through props:

```js
// useOverlayPosition.mjs:106-111
overlay.style.top = ''; overlay.style.bottom = ''; overlay.style.left = ''; overlay.style.right = '';
Object.keys(position.position).forEach((key) => overlay.style[key] = position.position[key] + 'px');
overlay.style.maxHeight = position.maxHeight != null ? position.maxHeight + 'px' : '';
```

In Chrome an open tooltip carries `style="left: 388.5px; bottom: 338px; max-height: 250px"`. `position: absolute` comes from `.base`. jsdom doesn't run layout, which is why the attribute looks empty there.

So Tooltip never received React Aria's inline z-index in the first place, and applying `overlayProps.style` directly would reintroduce `zIndex: 100000`.

packages/components/src/components/Modal/Modal.stories.tsx
KamilEmeleev: Can't reproduce this in a browser. Ran it on the PR's own preview build:

https://react-koobiq-next--prs-480-rz1v4tab.web.app/iframe.html?id=components-modal--stacking&viewMode=story

Press the button, open the outer dropdown, wait for the modal. Once the modal is up:

```js
document.querySelectorAll('[role="listbox"]').length  // 1
document.querySelectorAll('[role="dialog"]').length   // 1
getComputedStyle(popover).opacity                     // "1"
popover.getAttribute('data-transition')               // "entered"
```

The dropdown stays open and visible under the modal. Focus does move into the dialog, it just doesn't dismiss the popover.

jsdom handles focus containment and blur differently, which is probably where the 1 → 0 comes from. Leaving the story as is — happy to revisit if it reproduces in Chrome for you.

packages/components/src/components/Popover/Popover.module.css
KamilEmeleev: Checked this one and it turns out not to be a problem. While a popover is open React Aria's `ariaHideOutside` marks the underlay `inert`, so it's skipped by hit testing — `elementFromPoint` over it returns `body`, both before and after this PR. It wasn't blocking clicks either way, and dismissal runs on document-level listeners as you noted.

Separate thing I noticed while looking: we render the underlay unconditionally, while RAC only renders it for `!isNonModal`. That's worth a ticket on its own, and raising its z-index makes more sense together with that gate.

packages/components/src/components/Popover/PopoverInner.tsx
KamilEmeleev: We'd rather keep this in the components. The layer a component paints on is a component-level decision, not something the primitives layer should be deciding for everyone.

In practice `useOverlayPosition` is called in exactly two places: `PopoverInner`, which is the single entry point for all ten popover-based components, and `Tooltip`, where `useTooltip` filters `style` out so the inline value never lands anyway.

packages/components/src/components/Popover/Popover.module.css
KamilEmeleev: You're right that this changes: popovers used to sit at 100000, above the toast layer.

Leaving it for now though. `action` on a toast is normally a button or a link, and a dropdown opened from inside one is rare enough that we'd rather not carry the fix here — it needs the popover portaled into the toast region, which is a different change. Happy to pick it up if it actually comes up.

packages/components/src/components/Modal/Modal.mdx
KamilEmeleev: Agreed, a shared page under `docs/` is the right home for this. Leaving it out of this PR — will do it separately so the layer table, the DOM-order rule, the toast exception and the `portalContainer` caveat land in one place with links from the component pages.

packages/components/src/components/Modal/Modal.stories.tsx
KamilEmeleev: To be precise: the `status:updated` tag is gone from Modal, and we're not adding one to Popover either. No component gets tagged in this PR.

lskramarov (approved):

PR #481, opened 2026-09-10
fix(FormField): unify addon and action states in readonly and disabled (DS-5292)

KamilEmeleev:
### Notes

- The deprecated `Select` is intentionally left as is: it has no `isReadOnly` support, so it can't follow the new contract; a disabled `Select` still hides its clear button.

## Summary by CodeRabbit

- **New Features**
  - Improved read-only support across form controls, including date/time fields, selects, autocomplete, tag inputs, and tree selectors.
  - Read-only controls now consistently expose their state and prevent editing, clearing, removal, or opening interactive overlays.
  - Clear and remove buttons remain visible but disabled when controls are disabled or read-only.
  - Added form examples with disabled and read-only state toggles.

- **Bug Fixes**
  - Prevented actions such as clearing, tag removal, and time-range selection while read-only.

### DIFF
--- a/packages/components/src/components/Form/Form.stories.tsx
+++ b/packages/components/src/components/Form/Form.stories.tsx
@@ -1,5 +1,6 @@
 import { type FormEvent, useRef, useState } from 'react';
 
+import { Time, parseDate } from '@internationalized/date';
 import type { Meta, StoryObj } from '@storybook/react';
 
 import { Alert } from '../Alert';
@@ -16,11 +17,12 @@ import { InputNumber } from '../InputNumber';
 import { spacing } from '../layout';
 import { Radio, RadioGroup } from '../RadioGroup';
 import { SearchInput } from '../SearchInput';
-import { Select } from '../Select';
+import { SelectNext } from '../SelectNext';
 import { TagInput } from '../TagInput';
 import { Textarea } from '../Textarea';
 import { TimePicker } from '../TimePicker';
 import { TimeRange } from '../TimeRange';
+import { Toggle } from '../Toggle';
 import { TreeSelect } from '../TreeSelect';
 import { Typography } from '../Typography';
 
@@ -430,6 +432,9 @@ export const FormFields: Story = {
   },
   name: 'All form fields',
   render: function Render() {
+    const [isDisabled, setDisabled] = useState(false);
+    const [isReadOnly, setReadOnly] = useState(false);
+
     const tags = useListData<{ id: string; name: string }>({
       initialItems: [{ id: 'react', name: 'React' }],
     });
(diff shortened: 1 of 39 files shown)


rmnturov:
Я вообще колебался, скрывать всякие кнопки или оставлять в ридонли
Давай так оставим, как у тебя

Плейсхолдеры давай по умолчанию скрывать в обоих состояния

Ели они влияют на отображение (например, как в tag input), на общие габариты, то можно им прозрачность 0 делать

Плейсхолдеры
1) они не нужны в этих состояниях (ничего ввести нельзя)
2) были баги, что люди их путали с введенными значениями (неудачные формулировки были)

Я хочу сделать это опцией, которую можно выключить (выключить скрытие в ридонли и disabled)

rmnturov:
А по кнопкам, я готов выпустить в релиз с таким решением (оставлять бледными все кнопки, степперы). Мне нравится, что это решение не заставит контрол изменить свой размер.

KamilEmeleev:
@rmnturov https://react-koobiq-next--prs-481-b95r9lmf.web.app/?path=/story/components-form--form-fields здесь можно потыкать поля формы в состояниях read-only и disabled

packages/components/src/components/TimeRange/TimeRange.tsx
```diff
@@ -226,12 +232,28 @@ export function TimeRangeRender<T extends DateValue>(
...
+      ? {
+          ...overlayState,
+          open() {
+            return undefined;
+          },
+          toggle() {
+            return undefined;
+          },
+        }
+      : overlayState;
```
lskramarov, line 259: **Neutralizing `open`/`toggle` does not cover every entry point.**

The comment says the editor "cannot be opened when the component is disabled or read-only", but only `open()` and `toggle()` are replaced. `overlayState.isOpen` still comes straight from `slotProps.popover.defaultOpen` / `slotProps.popover.isOpen`, and `state.setOpen` is passed through unchanged to `<Popover onOpenChange={state.setOpen}>`.

```tsx
<TimeRange isReadOnly slotProps={{ popover: { defaultOpen: true } }}>
  <TimeRange.Field label="Period" />
</TimeRange>
```

renders the editor open on a read-only control. A controlled `slotProps.popover.isOpen` behaves the same way. Either fold those inputs into the guard (e.g. force `isOpen: false` when `isDisabled || isReadOnly`) or soften the comment so it does not claim full coverage.

packages/components/src/components/TimeRange/TimeRange.tsx
```diff
@@ -338,30 +360,33 @@ export function TimeRangeRender<T extends DateValue>(
...
+            triggerProps.onPress?.(event);
           }}
-          isDisabled={disabled}
+          isDisabled={isDisabled}
           isPressed={state.isOpen}
           ref={mergeRefs(ref, triggerRef)}
         >
-          <ButtonContext.Provider value={{ isDisabled: disabled }}>
+          <ButtonContext.Provider
+            value={{ isDisabled: isDisabled || isReadOnly }}
```
lskramarov, line 392: **(about line 424, just below this hunk)** `<TimeRangeEditor ... isDisabled={isReadOnly} />`

The editor is handed *only* the read-only flag — the component's own `isDisabled` never reaches it. In a PR that unifies read-only and disabled, this is the one spot where they stay split:

* a **disabled** TimeRange whose popover is opened via `slotProps.popover.defaultOpen`/`isOpen` shows a fully interactive editor;
* in **read-only** the fields are disabled, but the Apply button is gated only by `isDisabled={isDraftInvalid}` (line 431), so pressing it runs `handleApply()` → `setCommitted(...)` → `onChange` fires on a read-only control.

Both the editor and the Apply button should take `isDisabled || isReadOnly` into account.

packages/components/src/components/TimeRange/TimeRange.tsx
```diff
@@ -338,30 +360,33 @@ export function TimeRangeRender<T extends DateValue>(
...
+        value={{ formattedValue, isEmpty, isDisabled, isReadOnly, groupRef }}
       >
         <PressResponder
           {...triggerProps}
           {...{
             role: 'button',
             'aria-haspopup': 'dialog',
-            'aria-disabled': disabled || undefined,
+            'aria-disabled': isDisabled || undefined,
+            'aria-readonly': isReadOnly || undefined,
```
lskramarov: **`aria-readonly` is not allowed on `role="button"`.**

`aria-readonly` is not a global ARIA property, and the `button` role does not support it (ARIA 1.2 supports it on `checkbox`, `combobox`, `grid`, `gridcell`, `listbox`, `radiogroup`, `slider`, `spinbutton`, `textbox`, `searchbox`, `switch`, `treegrid`). `PressResponder` forwards these DOM props into `usePress`'s `pressProps`, so the trigger ends up as `role="button" aria-haspopup="dialog" aria-readonly="true"` — axe-core's `aria-allowed-attr` rule fails and AT ignores the attribute, so read-only is still not conveyed.

Every other control touched by this PR signals the state with `data-readonly` (TreeSelect, SelectNext, Input, TagInput, FormFieldControlGroup). Same here would be consistent:

```suggestion
            'data-readonly': isReadOnly || undefined,
```

packages/primitives/src/behaviors/useTagField.ts
```diff
@@ -581,8 +584,9 @@ export function useTagField<T extends object>(
   const tagListProps = {
     state,
     isDisabled,
+    isReadOnly,
     tabIndex: -1,
-    onRemove: isReadOnly ? undefined : handleRemove,
+    onRemove: handleRemove,
```
lskramarov, line 589: **Read-only tag lists now advertise — and swallow — the Delete/Backspace shortcut.**

Dropping `isReadOnly ? undefined : handleRemove` makes `allowsRemoving = !!onRemove` (`useTagListItem.ts:114`) true in read-only. Two consequences there:

1. `useTagListItem.ts:135-140` — the description is gated on `allowsRemoving && !isDisabled && (modality === 'keyboard' || 'virtual')`, so a focused tag in a **read-only** TagInput now announces *"Press Delete or Backspace to remove."* for a shortcut that does nothing.
2. `useTagListItem.ts:238-246` — Backspace/Delete now takes the `allowsRemoving` branch, which calls `event.preventDefault()` and does **not** `continuePropagation()`, so the key is swallowed before `handleRemove` early-returns.

Making the remove button visible-but-disabled is the goal, but `allowsRemoving` is currently doing double duty as "render the button" and "removal is possible". Passing the read-only state down (`useTagListItem` already receives it indirectly) so it can suppress the description and let the key event propagate would keep the visual change without the AT regression.

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
```diff
@@ -94,7 +96,11 @@ export function TagItem<T extends object>(props: TagItemProps<T>) {
           TagRemoveButtonProps,
           TagRemoveButtonProps | undefined,
         ]
-      >({ tabIndex: -1 }, removeButtonPropsAria, slotProps?.removeIcon)
+      >(
+        { tabIndex: -1, isDisabled: isReadOnly || isDisabled },
```
lskramarov: **Merge order lets an item-level `isDisabled={false}` re-enable the remove button in read-only.**

`mergeProps` resolves conflicts as `b !== undefined ? b : a` (`@react-aria/utils/mergeProps`), and the second argument `removeButtonPropsAria` carries its own `isDisabled` (`useTagListItem.ts:288`). That value is `isDisabledProp || selectionManager.isDisabled(key) || itemProps?.isDisabled` — the last operand wins when all are falsy, so it is `undefined` in the common case but a **defined `false`** as soon as a consumer writes it out:

```tsx
<TagInput isReadOnly>
  <TagInput.Tag isDisabled={false}>a</TagInput.Tag>
</TagInput>
```

→ `undefined || false || false` → `false`, which overwrites the `true` from `isReadOnly || isDisabled` and renders an enabled remove button on a read-only tag. Removal is still blocked by `handleRemove`, so this is a visual/AT-state bug rather than data loss.

Moving the read-only flag into its own object *after* `removeButtonPropsAria` (and before `slotProps?.removeIcon`, so the consumer escape hatch still wins) makes it immune to the item-level value:

```ts
mergeProps(
  { tabIndex: -1 },
  removeButtonPropsAria,
  { isDisabled: isReadOnly || isDisabled },
  slotProps?.removeIcon
)
```

packages/components/src/components/Input/Input.tsx
```diff
@@ -113,7 +113,7 @@ export const Input = forwardRef<InputRef, InputProps>((props, ref) => {
     <FormField as={TextField} {...rootProps}>
       {({ isInvalid, isRequired, isDisabled, state }) => {
         const hasValue = state.value !== '';
-        const clearButtonIsHidden = !hasValue || isDisabled || isReadOnly;
+        const clearButtonIsHidden = !hasValue;
```
lskramarov, line 116: **The clear action itself is still unguarded for Input — only the rendered `disabled` attribute protects it now.**

This PR consistently added a state-level guard everywhere the clear button became visible-but-disabled: `Autocomplete.tsx:144`, `SelectNext/Select.tsx:148`, `TreeSelect.tsx:245`, `useTagField.ts:221/268` all early-return on `isReadOnly`/`isDisabled`. Input is the exception — its clear handler lives in the primitive:

```ts
// packages/primitives/src/components/TextField/TextField.tsx:89
const handleClear = useCallback(() => {
  setInputValue('');
  onClear?.();
  inputRef?.current?.focus();
}, [setInputValue, onClear]);
```

It has no `isReadOnly`/`isDisabled` check, even though the sibling Escape path right below it (line 107) already checks `!isReadOnly`. Now that `clearButtonIsHidden` no longer folds in the two states, the only thing stopping it is the `disabled` attribute that `FormFieldClearButton` puts on the button via `useFormFieldControlGroup()` — which returns `{}` for anything rendered outside a `FormFieldControlGroup`, and does nothing for any other caller of the `clear-button` `ButtonContext` slot.

Adding `if (isDisabled || isReadOnly) return;` to that `handleClear` would put the guard at the same depth as the other five.

packages/components/src/components/TimeRange/TimeRange.tsx
```diff
@@ -338,30 +360,33 @@ export function TimeRangeRender<T extends DateValue>(
...
             'data-testid': testId,
-            tabIndex: disabled ? -1 : undefined,
+            tabIndex: isDisabled ? -1 : undefined,
```
lskramarov, line 380: **Read-only is now keyboard-focusable and press-enabled, but the test that covered the old behaviour wasn't updated.**

`tabIndex: disabled ? -1 : undefined` → `tabIndex: isDisabled ? -1 : undefined` and `isDisabled={disabled}` → `isDisabled={isDisabled}` deliberately put the read-only trigger back in the tab order (matching TreeSelect/SelectNext). That change has no test, and the existing one still claims the opposite:

```ts
// TimeRange.test.tsx:230-259
'should block a %s when the root is disabled or read-only'
  ...
  await userEvent.tab();
  expect(getTrigger()).not.toHaveFocus();
```

That assertion now only passes incidentally — the fixture renders a single focusable element, so `tab()` wraps focus to `document.body` after the preceding `click()` already focused the trigger. Worth splitting the read-only case out and asserting the new contract (focusable, `aria-expanded="false"`, dialog does not open), the way `TreeSelect.test.tsx:560` does with `expect(getControl()).toHaveAttribute('tabindex', '0')`.

packages/components/src/components/TimeRange/components/TimeRangeField.tsx
```diff
@@ -119,6 +125,7 @@ export const TimeRangeField = forwardRef<
   const groupProps = mergeProps<(FormFieldControlGroupProps | undefined)[]>(
     {
       isDisabled,
+      isReadOnly,
```
lskramarov, line 128: **(about `onMouseDown` at line 131-142)** That guard still reads `if (isDisabled || ...) return;`, but `isDisabled` no longer folds in read-only — the context used to carry `isDisabled || isReadOnly` (`TimeRange.tsx`, the removed `const disabled = isDisabled || isReadOnly`). So the guard silently narrowed.

Mousedown on the field's padding in read-only now runs `controlRef.current?.focus()` **and** `controlRef.current?.click()`, dispatching a synthetic DOM click on the trigger. The dialog stays closed because `onPress` is guarded upstream, but any consumer handler supplied through `slotProps.control.onClick` now fires on a read-only control where it previously did not. If the focus part is intended (read-only is focusable now), the `click()` probably still wants `if (isReadOnly) return;` before it.

packages/components/src/components/FormField/FormFieldAddon/FormFieldAddon.module.css
```diff
@@ -13,6 +13,10 @@
   --form-field-addon-color: var(--kbq-icon-error);
 }
 
+.readonly {
+  --form-field-addon-color: var(--kbq-states-icon-disabled);
+}
```
lskramarov: `.readonly` is a byte-for-byte copy of `.disabled` two rules below, and because it is declared *after* `.invalid` at equal specificity it also silently outranks it — that precedence is an artefact of source order rather than something the file states.

Concretely: an invalid, read-only Input with a `startAddon` gets `clsx(s.base, s.start, s.invalid, s.readonly)` and resolves `--form-field-addon-color` to `--kbq-states-icon-disabled`, losing the error colour. That is probably the intent (it matches `.disabled`), but folding the two selectors into one rule removes the duplication and makes the ordering deliberate:

```suggestion
.readonly,
.disabled {
  --form-field-addon-color: var(--kbq-states-icon-disabled);
}
```
(and drop the now-redundant `.disabled` rule below)

packages/components/src/components/TimeRange/TimeRange.tsx
```diff
@@ -226,12 +232,28 @@ export function TimeRangeRender<T extends DateValue>(
...
     defaultOpen: slotProps?.popover?.defaultOpen,
     onOpenChange: slotProps?.popover?.onOpenChange,
   });
 
+  // The editor cannot be opened when the component is disabled or read-only,
+  // so the trigger's own state is neutralized instead of guarding every entry
+  // point (same approach as SelectNext).
+  const state =
+    isDisabled || isReadOnly
+      ? {
```
lskramarov, line 246: This is the **third** hand-rolled copy of "spread the state, stub out `open`/`toggle`":

* `SelectNext/Select.tsx:441-451` — guards `isReadOnly`
* `packages/primitives/src/behaviors/useTreeSelectState.ts:209-218` — guards `isReadOnly` plus empty collection
* here — guards `isDisabled || isReadOnly`

Three slightly different variants, none of which covers `setOpen`, and the comment points at SelectNext instead of sharing the code. A small helper in primitives (e.g. `blockOverlayTrigger(state, isBlocked)`) would fix all three at once and could also be memoised — as written this object is rebuilt on every render whenever the component is disabled or read-only, which is the only case where the identity of `state` churns.

packages/components/src/components/Form/Form.stories.tsx
```diff
@@ -16,11 +17,12 @@ import { InputNumber } from '../InputNumber';
 import { spacing } from '../layout';
 import { Radio, RadioGroup } from '../RadioGroup';
 import { SearchInput } from '../SearchInput';
-import { Select } from '../Select';
+import { SelectNext } from '../SelectNext';
```
lskramarov, line 20: Swapping the showcase from `Select` to `SelectNext` in the same commit that adds the `isDisabled`/`isReadOnly` toggles means the deprecated `Select` is no longer exercised by the one story where the divergence would be visible.

`packages/components/src/components/Select/Select.tsx:86` still computes:

```ts
const clearButtonIsHidden = isDisabled || !state.selectedItems;
```

so a disabled `Select` **hides** its clear button, while a disabled `SelectNext`, `Input`, `SearchInput`, `TagInput`, `Autocomplete` and `TreeSelect` now all show a disabled one. Per `docs/9-deprecation-strategy.mdx` the deprecated component coexists until the next major, so consumers still on `Select` will see two conventions side by side in the same form. Either bring `Select` along (it is a one-line change there) or note in the PR why it is deliberately left behind.

packages/components/src/components/FormField/FormFieldControlGroup/FormFieldControlGroup.test.tsx
```diff
@@ -28,4 +28,40 @@ describe('FormFieldControlGroup', () => {
...
+        endAddon={<span>end</span>}
+      >
+        <input data-testid="input" />
+      </FormFieldControlGroup>
+    );
+
+    expect(getRoot()).toHaveAttribute('data-readonly', 'true');
+
+    ['start', 'end'].forEach((placement) => {
+      expect(screen.getByTestId(`field-addon-${placement}`)).toHaveAttribute(
```
lskramarov: Both new tests assert only `data-readonly`, but nothing in the repo styles `[data-readonly]` — `grep -rn "readonly" packages/components/src --include=*.css` returns exactly one hit, `.readonly` in `FormFieldAddon.module.css`. The visual change this PR ships therefore comes entirely from the `s.readonly` **class**, and no test covers it.

Deleting `isReadOnly && s.readonly` from `FormFieldAddon.tsx:38` would leave every read-only addon in the normal colour while both of these tests still pass, because `data-readonly` is set independently on the line below. One extra assertion closes the gap:

```ts
expect(screen.getByTestId('field-addon-end')).toHaveClass(addonStyles.readonly);
```

packages/components/src/components/TimeRange/types.ts
```diff
@@ -99,6 +99,7 @@ export type TimeRangeFieldProps = {
   isRequired?: boolean;
   isInvalid?: boolean;
   isDisabled?: boolean;
+  isReadOnly?: boolean;
```
lskramarov, line 107: `AGENTS.md` (Component Anatomy): *"every prop has a JSDoc description (`@default` where relevant) — Storybook's Props table is generated from those comments."*

This is a public prop (it lands in `tools/public_api_guard/components/TimeRange.api.md`) and has no description, so `TimeRange.Field`'s Props table will show an empty cell. The other two `isReadOnly` props added in this PR both got one — `FormFieldControlGroupProps` (with `@selector [data-readonly]`) and `TagListInnerProps`.

Worth mentioning that the parent wins, since `TimeRangeField.tsx:76` puts `contextReadOnly` first in the `||` chain:

```suggestion
  /**
   * Whether the field can be focused but not changed by the user.
   * A read-only `TimeRange` parent overrides this value.
   * @selector [data-readonly]
   */
  isReadOnly?: boolean;
```

NikGurev (approved):

packages/primitives/src/behaviors/useTagField.ts
KamilEmeleev: Fixed. `useTagListItem` now accepts `isReadOnly`: the remove button still renders (disabled), but a read-only tag no longer announces the Delete/Backspace hint and lets the key propagate like a non-removable tag. Regression test added in `TagInput.test.tsx`.

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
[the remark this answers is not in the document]
KamilEmeleev: Fixed together with the `useTagField.ts` thread: read-only tags stay focusable, Delete/Backspace removal is off, regression test added.

packages/components/src/components/TimeRange/TimeRange.tsx
KamilEmeleev: Fixed: the editor and the Apply button both take `isDisabled || isReadOnly` now.

packages/components/src/components/TimeRange/TimeRange.tsx
KamilEmeleev: Fixed, replaced with `data-readonly` as suggested.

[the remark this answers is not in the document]
KamilEmeleev: Fixed, replaced with `data-readonly`.

packages/components/src/components/TimeRange/TimeRange.tsx
KamilEmeleev: Fixed: while disabled or read-only, `isOpen` is masked to `false` and `setOpen`/`open`/`toggle` are no-ops, so `slotProps.popover.defaultOpen`/`isOpen` can't open the editor either. Comment updated, test added.

packages/components/src/components/TimeRange/components/TimeRangeField.tsx
KamilEmeleev: Fixed: a read-only field still takes focus on mousedown but no longer dispatches the synthetic click.

packages/components/src/components/Input/Input.tsx
KamilEmeleev: Fixed: `handleClear` in the `TextField` primitive now returns early when disabled or read-only.

packages/components/src/components/Form/Form.stories.tsx
KamilEmeleev: Left intentionally: `Select` is deprecated and has no `isReadOnly` support at all, so it can't follow the new contract anyway. Noted this in the PR description.

packages/components/src/components/TagList/components/TagItem/TagItem.tsx
KamilEmeleev: Fixed: the disabled state is merged after React Aria's props and before `slotProps.removeIcon`, as suggested.

packages/components/src/components/TimeRange/TimeRange.tsx
KamilEmeleev: Fixed: the test is split. Disabled triggers are checked with Tab from a fresh render, so the assertion no longer passes by accident. A read-only `TimeRange.Field` is asserted focusable, `aria-expanded="false"`, and Enter doesn't open the editor. Custom `Button` triggers are still disabled through `ButtonContext` in read-only (unchanged by this PR), so for custom triggers the test only checks that the editor doesn't open.

packages/components/src/components/TimeRange/TimeRange.tsx
```diff
@@ -162,15 +163,20 @@ export function TimeRangeRender<T extends DateValue>(
...
     style,
     'data-testid': testId,
     slotProps,
     ...other
   } = props;
 
+  const { isDisabled: formIsDisabled, isReadOnly: formIsReadOnly } = useForm();
+
+  const isDisabled = isDisabledProp ?? formIsDisabled;
+  const isReadOnly = isReadOnlyProp ?? formIsReadOnly;
```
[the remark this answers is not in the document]
KamilEmeleev: Added `should inherit isDisabled/isReadOnly from a parent Form`.

packages/components/src/components/FormField/FormFieldControlGroup/FormFieldControlGroup.test.tsx
KamilEmeleev: Added the `toHaveClass(addonStyles.readonly)` assertion.

packages/components/src/components/FormField/FormFieldAddon/FormFieldAddon.module.css
KamilEmeleev: Done, merged into one `.readonly, .disabled` rule.

packages/components/src/components/TimeRange/types.ts
KamilEmeleev: Added, with your wording.

packages/components/src/components/TimeRange/TimeRange.tsx
KamilEmeleev: Agreed. Moved to a separate task: a shared helper in primitives for SelectNext, `useTreeSelectState` and TimeRange, covering `setOpen` and memoization.

packages/components/src/components/SelectNext/Select.tsx
```diff
@@ -184,7 +183,7 @@ function SelectInner<T extends object, M extends SelectionMode = 'single'>({
     'data-disabled': isDisabled || undefined,
     'data-readonly': isReadOnly || undefined,
     'data-required': isRequired || undefined,
-    className: clsx(s.base, className),
+    className,
```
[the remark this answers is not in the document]
KamilEmeleev: Not needed: `.base` no longer exists in `Select.module.css`. Its only rule was the read-only chevron color, which moved into `FormFieldAddon`. `s.base` was `undefined`, so no styles are lost.

KamilEmeleev:
Кнопки оставляем как есть. Скрытие плейсхолдеров в readonly и disabled (с `opacity: 0`, где плейсхолдер влияет на размер, и опцией, чтобы вернуть показ) вынес в отдельную задачу: это новое поведение и новый проп, в этот PR не берём.

packages/components/src/components/TimeRange/TimeRange.tsx
KamilEmeleev: Tracked in DS-5563.

KamilEmeleev:
Задача на скрытие плейсхолдеров: DS-5564.

rmnturov (approved):

lskramarov (approved):

PR #482, opened 2026-09-10
feat(components): add `FileTrigger` component (DS-5421)

KamilEmeleev:
## Summary by CodeRabbit

- **New Features**
  - Added the `FileTrigger` component for opening the system file picker from buttons, links, and other pressable elements.
  - Supports multiple file selection, accepted file types, directory selection, camera selection on mobile, disabled states, programmatic opening, and accessibility use cases.
  - Added comprehensive Storybook examples and public API documentation.

- **Documentation**
  - Documented usage, supported triggers, configuration options, accessibility guidance, and the distinction from `FileUpload`.


packages/components/src/components/FileTrigger/FileTrigger.tsx
```diff
@@ -0,0 +1,23 @@
...
+/**
+ * FileTrigger opens the system file dialog from any pressable component.
+ * It renders no markup of its own, apart from a hidden file input.
+ */
+export const FileTrigger = forwardRef<FileTriggerRef, FileTriggerProps>(
+  (props, ref) => {
+    const { accept, ...other } = props;
+
+    return (
+      <FileTriggerPrimitive ref={ref} acceptedFileTypes={accept} {...other} />
```
lskramarov, line 18: **`acceptDirectory` is forwarded without `multiple`, unlike `FileUpload.Trigger`**

Rendering `<FileTrigger acceptDirectory>` (the `Directory` story) produces `<input type="file" webkitdirectory>` with `multiple` absent — I verified this by rendering it in jsdom.

`FileUpload` deliberately does the opposite. `FileUploadTrigger.tsx:61`:

```tsx
allowsMultiple={allowsMultiple || isDirectory}
```

and `FileUpload.test.tsx:302` locks that invariant in:

```
it('allows the folder picker to return every file in single mode', ...)
  expect(input).toHaveAttribute('webkitdirectory');
  expect(input).toHaveAttribute('multiple');
```

So the same folder, picked through `FileTrigger`, can come back as a single file in the browsers that invariant was written for, while `FileUpload` returns every file — two sibling components in one library behaving differently for the same feature. Either pair `acceptDirectory` with `multiple` here the way `FileUpload` does, or document explicitly why `FileTrigger` should not.

packages/components/src/components/FileTrigger/FileTrigger.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+      <FileTrigger {...baseProps} ref={ref}>
+        <Button>Choose</Button>
+      </FileTrigger>
+    );
+
+    const click = spyOnFileDialog();
+
+    ref.current?.click();
+
+    expect(click).toHaveBeenCalledTimes(1);
```
lskramarov, line 192: **This test is tautological — it can never fail.**

`spyOnFileDialog()` does `vi.spyOn(getFileInput(), 'click')`, and `ref.current` *is* that same input. So line 207 invokes the spy directly and line 209 asserts the spy recorded its own call; the component is not involved at all.

Verified by mutation: replacing `FileTrigger`'s body with a bare `<input type="file" ref={ref} />` (no `PressResponder`, no prop forwarding whatsoever) fails **8 of the 12** tests in this file — but this one still passes.

`should accept the ref pointing to the file input` (line 47) already covers ref identity. To actually cover the programmatic path, spy on `HTMLInputElement.prototype.click` before render and assert the RAC input received it — otherwise the test is better dropped.

packages/components/src/components/FileTrigger/FileTrigger.stories.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+    return (
+      <FlexBox direction="column" gap="s" alignItems="center">
+        <FlexBox gap="s">
+          <FileTrigger {...args} ref={inputRef} onSelect={onSelect}>
+            <Button>Choose a file</Button>
+          </FileTrigger>
+          <Button
+            variant="fade-contrast-filled"
+            onPress={() => inputRef.current?.click()}
```
lskramarov: **The programmatic open never resets `value`, so re-picking the same file does nothing.**

In a real browser: press *Open from elsewhere*, pick `a.txt` → `onSelect` fires. Press it again and pick `a.txt` again → `input.value` is unchanged, no `change` event fires, `onSelect` never runs, and the demo looks frozen.

React Aria's own press handler guards against exactly this (`FileTrigger.mjs`):

```js
if (inputRef.current?.value) inputRef.current.value = '';
inputRef.current?.click();
```

and `FileTrigger.mdx:92` tells readers to do `ref.current.value = ''` — but the one story that demonstrates programmatic opening is the place that doesn't.

```suggestion
            onPress={() => {
              if (!inputRef.current) return;

              inputRef.current.value = '';
              inputRef.current.click();
            }}
```

packages/components/src/components/FileTrigger/types.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export type FileTriggerPropDefaultCamera =
+  (typeof fileTriggerPropDefaultCamera)[number];
+
+export type FileTriggerProps = {
+  /**
+   * The trigger that opens the file dialog. Any pressable component works
+   * out of the box: `Button`, `IconButton`, `Link`.
+   */
+  children?: ReactNode;
```
lskramarov: **`children` should be required.**

`<FileTrigger onSelect={fn} />` compiles cleanly today, renders a hidden input nothing can open, and logs React Aria's `A PressResponder was rendered without a pressable child`. `FileTrigger.mdx:47-50` documents that footgun rather than preventing it.

That matters more now that `tools/vitest/setupTests.ts` runs `failOnConsole({ shouldFailOnWarn: true, shouldFailOnError: true })` — a consumer who makes this mistake gets an opaque suite failure instead of a compile error.

```suggestion
  children: ReactNode;
```

packages/components/src/components/FileTrigger/FileTrigger.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+      <FileTrigger {...baseProps}>
+        <Button isDisabled>Choose</Button>
+      </FileTrigger>
+    );
+
+    const click = spyOnFileDialog();
+
+    await user.click(screen.getByRole('button'));
+
+    expect(click).not.toHaveBeenCalled();
```
lskramarov, line 166: **This assertion is only a negative, so it passes vacuously.**

Same mutation as above: with `FileTrigger` replaced by a bare `<input type="file" />` that has no `PressResponder` at all, this test still passes — nothing ever opens the dialog, so `expect(click).not.toHaveBeenCalled()` holds for the wrong reason. It cannot distinguish "`isDisabled` blocks the dialog" from "the press wiring is gone".

Pair it with a positive in the same render (an enabled sibling trigger that *does* open), or additionally assert the rendered button actually carries the disabled state.

packages/components/src/components/FileTrigger/types.ts
```diff
@@ -0,0 +1,38 @@
...
+  allowsMultiple?: boolean;
+  /**
+   * Whether a directory is selected instead of individual files.
+   * @default false
+   */
+  acceptDirectory?: boolean;
+  /** Which camera the capture mechanism uses. Mobile browsers only. */
+  defaultCamera?: FileTriggerPropDefaultCamera;
+  /** Unique identifier for testing purposes. */
+  'data-testid'?: string | number;
```
lskramarov, line 34: **`data-testid` lands on the hidden input, which is the opposite of `FileUpload.Trigger`.**

`filterDOMProps` forwards `data-*` onto the `display: none` file input, so `getByTestId('attach')` returns the input, not the trigger. A consumer who writes:

```tsx
<FileTrigger data-testid="attach"><Button>Attach</Button></FileTrigger>
// ...
await user.click(screen.getByTestId('attach'));
```

clicks the hidden input directly, bypasses the `PressResponder` entirely, and skips the `value` reset the press handler performs — the test passes while proving nothing about the trigger.

`FileUploadTrigger.tsx:74` puts `data-testid` on the visible `Link`, so the same prop name means two different targets in the same library. At minimum the JSDoc should say where it lands:

```suggestion
  /** Unique identifier for testing purposes. Set on the hidden file input. */
  'data-testid'?: string | number;
```

packages/components/src/components/FileTrigger/FileTrigger.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+    render(
+      <FileTrigger {...baseProps} onSelect={onSelect}>
+        <Button>Choose</Button>
+      </FileTrigger>
+    );
+
+    await user.upload(getFileInput(), file);
+
+    expect(onSelect).toHaveBeenCalledTimes(1);
+    expect(Array.from(onSelect.mock.calls[0][0])).toEqual([file]);
```
lskramarov: **Unguarded index — a regression here surfaces as an opaque `TypeError`.**

If prop forwarding ever regresses and `onSelect` is not called, this line evaluates `Array.from(undefined[0])` and the test dies with `TypeError: Cannot read properties of undefined (reading '0')`, with no indication of which expectation failed. Same pattern at lines 88 and 106 — and line 88 has no preceding `toHaveBeenCalled` guard at all.

The root `tsconfig.json` sets `noUncheckedIndexedAccess: true` (AGENTS.md: *"TypeScript is strict with `noUncheckedIndexedAccess` (indexing yields `T | undefined`)"*) and flags all three: `tsc -p tsconfig.json` reports TS2532 at 66, 88, 106. CI's `pnpm type-check` misses them only because `packages/components/tsconfig.json` omits that flag.

`expect(onSelect).toHaveBeenLastCalledWith(expect.anything())` — or asserting through `mock.lastCall` after a `toHaveBeenCalledTimes` guard — keeps the failure readable.

packages/components/src/components/FileTrigger/types.ts
```diff
@@ -0,0 +1,38 @@
...
+   * out of the box: `Button`, `IconButton`, `Link`.
+   */
+  children?: ReactNode;
+  /** Handler that is called when a user selects files in the dialog. */
+  onSelect?: (files: FileList | null) => void;
+  /**
+   * The file types the dialog offers, for example `['image/*', '.pdf']`.
+   * The browser only filters the dialog, it does not validate the result.
+   */
+  accept?: readonly string[];
```
lskramarov, line 20: **`accept` has two incompatible types across sibling components.**

Here it is `readonly string[]`; `FileUploadProps['accept']` (`FileUpload/types.ts:130`) is `string[]`.

A consumer factoring out `const ACCEPT = ['image/*', '.pdf'] as const` to share between the two gets it accepted by `FileTrigger` and rejected by `FileUpload` (`readonly` is not assignable to a mutable array). The only shape that works for both is `string[]`, which forces them to drop the `as const`.

`readonly string[]` is the better of the two (it matches RAC's `acceptedFileTypes?: ReadonlyArray<string>`) — worth widening `FileUploadProps['accept']` to match so the library is consistent.

packages/components/src/components/FileTrigger/FileTrigger.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+import { FileTrigger as FileTriggerPrimitive } from '@koobiq/react-primitives';
+
+import type { FileTriggerRef, FileTriggerProps } from './types';
+
+/**
+ * FileTrigger opens the system file dialog from any pressable component.
+ * It renders no markup of its own, apart from a hidden file input.
+ */
+export const FileTrigger = forwardRef<FileTriggerRef, FileTriggerProps>(
```
lskramarov, line 13: **Consider rebasing `FileUpload.Trigger` onto this component.**

`FileUpload/components/FileUploadTrigger/FileUploadTrigger.tsx:58-79` still imports `FileTrigger` from `@koobiq/react-primitives` and repeats what this file does: the `accept` → `acceptedFileTypes` rename, and the ref-points-at-the-hidden-input contract.

Two places in the same package now own the primitive-wrapping logic, so any fix has to be remembered twice — the `acceptDirectory` / `multiple` pairing I flagged on line 18 is the immediate example. Building `FileUploadTrigger` on the public `FileTrigger` would collapse the duplication and make the folder-mode invariant a single implementation detail.

packages/components/src/components/FileTrigger/FileTrigger.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+      >
+        <Button>Choose</Button>
+      </FileTrigger>
+    );
+
+    expect(getFileInput()).toHaveAttribute('accept', 'image/*,.pdf');
+
+    await user.upload(getFileInput(), text);
+
+    expect(Array.from(onSelect.mock.calls[0][0])).toEqual([text]);
```
lskramarov: **The upload half of this test exercises `user-event`, not `FileTrigger`.**

`applyAccept: false` (line 70) plus a mismatched `.txt` only demonstrates that `@testing-library/user-event`'s own flag skips its filtering step. The component never reads back what the browser filtered, so there is nothing here for it to get wrong.

The real assertion is line 82 (`toHaveAttribute('accept', 'image/*,.pdf')`), and `should call onSelect with the selected files` (line 52) already covers the `onSelect` path. Dropping lines 70-72 and 84-88 narrows the test to what it names and removes one of the three unguarded `mock.calls[0][0]` indexings.

packages/components/src/components/FileTrigger/FileTrigger.tsx
KamilEmeleev: Keeping the React Aria behavior here. Per the Entries API spec, picking a directory with `webkitdirectory` works "as if all files with that directory as an ancestor were selected" — `multiple` plays no part, and browsers return the whole tree without it:
https://wicg.github.io/entries-api/#dom-htmlinputelement-webkitdirectory

The only place the missing `multiple` shows up is `user-event`'s `upload()`, which keeps just the first file on a non-`multiple` input. So `FileTrigger` stays a 1:1 wrapper over the primitive.

packages/components/src/components/FileTrigger/types.ts
KamilEmeleev: Done — `children` is required now, API report updated.

packages/components/src/components/FileTrigger/FileTrigger.stories.tsx
KamilEmeleev: Applied your suggestion — the button resets `value` before opening the dialog.

packages/components/src/components/FileTrigger/types.ts
KamilEmeleev: Kept the input as the target on purpose: it's the element consumers need for `user.upload(screen.getByTestId(...), file)`, and the trigger takes its own `data-testid`. Updated the JSDoc to say where it lands.

packages/components/src/components/FileTrigger/types.ts
KamilEmeleev: Agreed that `readonly string[]` is the right shape. Widening `FileUploadProps['accept']` changes `FileUpload`, so it's tracked separately in DS-5561, together with rebasing `FileUpload.Trigger` onto `FileTrigger`.

packages/components/src/components/FileTrigger/FileTrigger.tsx
KamilEmeleev: Makes sense. It changes `FileUpload`, so it's tracked in DS-5561 together with the `accept` type.

packages/components/src/components/FileTrigger/FileTrigger.mdx
```diff
@@ -0,0 +1,101 @@
...
+<Alert>
+  `accept` is a hint for the dialog, not validation. A user can still drag any
+  file into the dialog or pick one through "All files", so check the result in
+  `onSelect`.
+</Alert>
+
+## Directories
+
+With `acceptDirectory`, the dialog selects a whole directory. Every file keeps
+its path in `webkitRelativePath`:
```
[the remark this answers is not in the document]
KamilEmeleev: Added a note to the Directories section: older browsers ignore `acceptDirectory` and open the regular dialog. The Camera section already says desktop browsers ignore `defaultCamera`.

packages/components/src/components/FileTrigger/FileTrigger.test.tsx
KamilEmeleev: Right — removed it. The ref identity test already covers the contract.

packages/components/src/components/FileTrigger/FileTrigger.test.tsx
KamilEmeleev: Added a positive control: re-rendering the same trigger without `isDisabled` opens the dialog.

packages/components/src/components/FileTrigger/FileTrigger.test.tsx
KamilEmeleev: Switched to a typed `vi.fn` and `mock.lastCall` behind a `toHaveBeenCalledTimes(1)` guard. `tsc -p tsconfig.json` from the root is clean for these files now.

packages/components/src/components/FileTrigger/FileTrigger.test.tsx
KamilEmeleev: Trimmed it to the attribute check.

lskramarov (approved):

PR #483, opened 2026-09-10
feat(Resizable): support single-axis sizing (DS-5552)

KamilEmeleev:
`size` and `defaultSize` now accept a single dimension, so an element can be managed in one axis and keep its CSS size in the other. Panels stretched to the full height of their container no longer get that height frozen in pixels.

An axis becomes managed once it is given in `size` / `defaultSize` or resized by a handle; the resize handlers keep reporting both dimensions, measuring an unmanaged axis once, when the resize starts.

Also adds `disableKeyboardResize` to `Resizable.Handle` for drag-only handles, and starts a resize from the size the handles announce instead of the element rect, which drifted while the element animated and is scaled by CSS transforms.

## Summary by CodeRabbit

* **New Features**
  * Resizable components now support managing width and height independently, allowing the other dimension to retain its CSS-defined size.
  * Added `disableKeyboardResize` for handles, disabling arrow-key resizing while preserving pointer dragging.
  * Handles become non-focusable by default when keyboard resizing is disabled, with `tabIndex` available to override this behavior.

* **Documentation**
  * Added guidance on one-dimensional sizing and keyboard resize controls.

### DIFF
--- a/packages/components/src/components/Resizable/Resizable.mdx
+++ b/packages/components/src/components/Resizable/Resizable.mdx
@@ -54,6 +54,23 @@ The measured size is fixed in pixels only after the first resize interaction.
 
 <Story of={Stories.IntrinsicSize} />
 
+### One dimension
+
+`size` and `defaultSize` accept a single dimension. The omitted one keeps its CSS size and is
+never written to the element, which suits elements sized by their layout in the other axis —
+a side panel stretched to the full height of its container, for example:
+
+```tsx
+<Resizable defaultSize={{ width: 320 }} minSize={{ width: 240 }}>
+  <Content />
+  <Resizable.Handle direction={[1, 0]} />
+</Resizable>
+```
+
+An axis becomes managed as soon as it's given in `size` / `defaultSize`, or resized by a handle.
+The `onResize`, `onResizeStart` and `onResizeEnd` handlers always report both dimensions,
+measured where an axis isn't managed.
+
 ## Direction
 
 `Resizable.Handle` accepts a physical `[x, y]` direction. Each axis supports `-1`, `0`, and `1`:
@@ -85,6 +102,9 @@ for geometry customization.
 Handles are focusable. Use the arrow keys to resize by one pixel, or hold <kbd>Shift</kbd> to resize
 by ten pixels. Edge handles expose the ARIA separator pattern; corner handles expose an accessible
 two-dimensional resize control.
+
+Set `disableKeyboardResize` on a handle to leave it drag-only. It also drops out of the tab order,
+unless `tabIndex` says otherwise.
 
 ## Disabled
 
--- a/packages/components/src/components/Resizable/Resizable.mdx
+++ b/packages/components/src/components/Resizable/Resizable.mdx
@@ -50,7 +50,7 @@ All values are measured in CSS pixels.
 ### Intrinsic size
 
 When both `size` and `defaultSize` are omitted, Resizable preserves its natural CSS size.
-The measured size is fixed in pixels only after the first resize interaction.
+The first resize interaction fixes the resized axes in pixels, the others keep their CSS size.
 
 <Story of={Stories.IntrinsicSize} />
 
@@ -58,18 +58,14 @@ The measured size is fixed in pixels only after the first resize interaction.
 
 `size` and `defaultSize` accept a single dimension. The omitted one keeps its CSS size and is
 never written to the element, which suits elements sized by their layout in the other axis —
-a side panel stretched to the full height of its container, for example:
-
-```tsx
-<Resizable defaultSize={{ width: 320 }} minSize={{ width: 240 }}>
-  <Content />
-  <Resizable.Handle direction={[1, 0]} />
-</Resizable>
-```
+a side panel stretched to the full height of its container, for example.
 
 An axis becomes managed as soon as it's given in `size` / `defaultSize`, or resized by a handle.
-The `onResize`, `onResizeStart` and `onResizeEnd` handlers always report both dimensions,
-measured where an axis isn't managed.
+The `onResize`, `onResizeStart` and `onResizeEnd` handlers always report both dimensions.
+An axis that isn't managed is measured once, when the resize starts. To control a single axis,
+take only that axis from the handler: passing the whole size back would manage the other one too.
+
+<Story of={Stories.OneDimension} />
 
 ## Direction
 
--- a/packages/components/src/components/Resizable/Resizable.test.tsx
+++ b/packages/components/src/components/Resizable/Resizable.test.tsx
@@ -1,4 +1,4 @@
-import { createRef, type ComponentProps } from 'react';
+import { createRef, useState, type ComponentProps } from 'react';
 
 import type * as ReactCore from '@koobiq/react-core';
 import { fireEvent, render, screen } from '@testing-library/react';
@@ -70,33 +70,6 @@ describe('Resizable', () => {
   beforeEach(() => {
     mocks.observedSize = { width: 300, height: 200 };
     vi.stubGlobal('PointerEvent', undefined);
-
-    vi.spyOn(HTMLElement.prototype, 'getBoundingClientRect').mockImplementation(
-      function getBoundingClientRect(this: HTMLElement) {
-        const width = Number.parseFloat(this.style.width);
-        const height = Number.parseFloat(this.style.height);
-
-        const resolvedWidth = Number.isNaN(width)
-          ? mocks.observedSize.width
-          : width;
-
-        const resolvedHeight = Number.isNaN(height)
-          ? mocks.observedSize.height
-          : height;
-
-        return {
-          x: 0,
-          y: 0,
-          top: 0,
-          left: 0,
-          right: resolvedWidth,
-          bottom: resolvedHeight,
-          width: resolvedWidth,
-          height: resolvedHeight,
-          toJSON: () => ({}),
-        };
-      }
-    );
   });
 
   afterEach(() => {
@@ -315,6 +288,28 @@ describe('Resizable', () => {
     expect(getRoot().style.height).toBe('');
   });
 
+  it('keeps the omitted axis unmanaged in controlled state', () => {
+    function ControlledWidth() {
+      const [width, setWidth] = useState(300);
+
+      return (
+        <Resizable
+          data-testid="resizable"
+          size={{ width }}
+          onResize={(size) => setWidth(size.width)}
+        >
+          <Resizable.Handle data-testid="handle" direction={[1, 0]} />
+        </Resizable>
+      );
+    }
+
+    render(<ControlledWidth />);
+    drag(20, 10);
+
+    expect(getRoot()).toHaveStyle({ width: '320px' });
+    expect(getRoot().style.height).toBe('');
+  });
+
   it('keeps the managed axis when another axis is resized', () => {
     render(
       <Resizable data-testid="resizable" defaultSize={{ width: 300 }}>
@@ -328,41 +323,32 @@ describe('Resizable', () => {
   });
 
   it('disables arrow key resizing without breaking dragging', () => {
-    renderResizable([1, 0], { defaultSize: { width: 300 } });
-
     render(
       <Resizable data-testid="resizable" defaultSize={{ width: 300 }}>
         <Resizable.Handle
-          data-testid="static-handle"
+          data-testid="handle"
           direction={[1, 0]}
           disableKeyboardResize
         />
       </Resizable>
     );
 
-    const handle = screen.getByTestId('static-handle');
-    const root = handle.parentElement as HTMLElement;
+    expect(getHandle()).toHaveAttribute('tabindex', '-1');
 
-    expect(handle).toHaveAttribute('tabindex', '-1');
+    fireEvent.keyDown(getHandle(), { key: 'ArrowRight' });
 
-    fireEvent.keyDown(handle, { key: 'ArrowRight' });
+    expect(getRoot()).toHaveStyle({ width: '300px' });
 
-    expect(root).toHaveStyle({ width: '300px' });
+    drag(20, 0);
 
-    fireEvent.mouseDown(handle, { button: 0, clientX: 10, clientY: 10 });
-    fireEvent.mouseMove(window, { button: 0, clientX: 30, clientY: 10 });
-    fireEvent.mouseUp(window, { button: 0, clientX: 30, clientY: 10 });
-
-    expect(root).toHaveStyle({ width: '320px' });
+    expect(getRoot()).toHaveStyle({ width: '320px' });
   });
 
   it('keeps the handle focusable when tabIndex is set explicitly', () => {
-    renderResizable([1, 0], { defaultSize: { width: 300 } });
-
     render(
       <Resizable data-testid="resizable" defaultSize={{ width: 300 }}>
         <Resizable.Handle
-          data-testid="static-handle"
+          data-testid="handle"
           direction={[1, 0]}
           tabIndex={0}
           disableKeyboardResize
@@ -370,10 +356,7 @@ describe('Resizable', () => {
       </Resizable>
     );
 
-    expect(screen.getByTestId('static-handle')).toHaveAttribute(
-      'tabindex',
-      '0'
-    );
+    expect(getHandle()).toHaveAttribute('tabindex', '0');
   });
 
   it('calls lifecycle callbacks with the snapshot and final size', () => {
@@ -529,6 +512,24 @@ describe('Resizable', () => {
     expect(getHandle()).not.toHaveAttribute('aria-valuenow');
     expect(getHandle()).not.toHaveAttribute('aria-orientation');
     expect(getHandle()).toHaveAttribute('aria-controls', getRoot().id);
+  });
+
+  it('does not advertise arrow keys when keyboard resizing is disabled', () => {
+    render(
+      <Resizable
+        data-testid="resizable"
+        defaultSize={{ width: 300, height: 200 }}
+      >
+        <Resizable.Handle
+          data-testid="handle"
+          direction={[1, 1]}
+          disableKeyboardResize
+        />
+      </Resizable>
+    );
+
+    expect(getHandle()).toHaveAttribute('role', 'button');
+    expect(getHandle()).not.toHaveAttribute('aria-keyshortcuts');
   });
 
   it('localizes default accessible labels', () => {
--- a/packages/components/src/components/Resizable/Resizable.test.tsx
+++ b/packages/components/src/components/Resizable/Resizable.test.tsx
@@ -280,7 +280,100 @@ describe('Resizable', () => {
 
     fireEvent.keyDown(getHandle(), { key: 'ArrowRight' });
 
-    expect(getRoot()).toHaveStyle({ width: '261px', height: '140px' });
+    // Only the resized axis becomes managed, the other one is left to CSS.
+    expect(getRoot()).toHaveStyle({ width: '261px' });
+    expect(getRoot().style.height).toBe('');
+  });
+
+  it('manages a single dimension when the size defines one axis', () => {
+    const onResize = vi.fn();
+
+    mocks.observedSize = { width: 300, height: 200 };
+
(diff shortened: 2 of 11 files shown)


packages/components/src/components/Resizable/hooks/useResizableState.ts
```diff
@@ -72,16 +72,40 @@ export const useResizableState = (
...
+
+      lastSizeRef.current = nextSize;
+
+      setManagedSize({
+        ...managedSizeRef.current,
+        ...(x !== 0 && { width: nextSize.width }),
+        ...(y !== 0 && { height: nextSize.height }),
+      });
+
+      onResize?.(nextSize);
```
lskramarov, line 101: **correctness** — `onResize` reports both axes, so a controlled single-axis Resizable re-freezes the unmanaged one.

`setManagedSize` just above correctly writes only the touched axes, but `onResize` gets the full two-axis `nextSize`. A controlled consumer feeding the callback back into `size` — the exact pattern the `Base` story and the `## State` section teach — therefore promotes the omitted axis to managed on the very first drag.

Verified against this branch:

```tsx
const [size, setSize] = useState<ResizableSizeConstraints>({ width: 300 });

<Resizable size={size} onResize={setSize}>
  <Resizable.Handle direction={[1, 0]} />
</Resizable>;
```

One `ArrowRight` press → `onResize({ width: 301, height: 200 })` → the root’s inline style becomes `width: 301px; height: 200px`. The height is now frozen in pixels — precisely what the PR description says single-axis sizing removes (“Panels stretched to the full height of their container no longer get that height frozen in pixels”). The feature works uncontrolled and breaks controlled.

Either report the managed shape (`ResizableSizeConstraints`) so it can be echoed back safely, or hand the merged value to `onResize` alongside the measured one.

packages/components/src/components/Resizable/hooks/useResizable.ts
```diff
@@ -56,18 +58,27 @@ export const useResizable = <T extends HTMLElement = HTMLElement>(
   } = useElementSize<T>({ box: 'border-box' });
 
   const observedSize = clampResizableSize({ width, height }, bounds);
-  const currentSize = managedSize ?? observedSize;
+
+  const currentSize = clampResizableSize(
```
lskramarov: **efficiency** — this rebuild defeats the `contextValue` memo on line 84.

`currentSize` is now always a fresh object literal, so the `useMemo` deps never match and a new context value is produced on every render. Previously `managedSize ?? observedSize` returned the stable state object whenever an axis was managed.

Measured with a context-identity probe (`defaultSize={{ width: 300, height: 200 }}`, three renders driven by an unrelated parent state change):

- this branch: `renders=3 uniqueContexts=3`
- with `const currentSize = managedSize ?? observedSize`: `renders=3 uniqueContexts=1`

Every `Resizable.Handle` then re-renders and rebuilds its `useMove` / `useHover` / `useFocusRing` props — 8× per render for the 8-handle layout in the stories, on the drag hot path.

Also note `observedSize` on line 60 is already clamped and gets clamped a second time here. Memoizing and clamping once fixes both:

```suggestion
  const currentSize = useMemo(
    () =>
      clampResizableSize(
        {
          width: managedSize?.width ?? width,
          height: managedSize?.height ?? height,
        },
        bounds
      ),
    [managedSize?.width, managedSize?.height, width, height, bounds]
  );
```

packages/components/src/components/Resizable/hooks/useResizableHandle.ts
```diff
@@ -70,7 +84,7 @@ export const useResizableHandle = (
     defaultLabel = t.format('resize height');
   }
 
-  let tabIndex = tabIndexProp ?? 0;
+  let tabIndex = tabIndexProp ?? (disableKeyboardResize ? -1 : 0);
```
lskramarov, line 85: **accessibility** — the handle leaves the tab order but keeps its widget ARIA.

Nothing below (lines 107–137) reacts to `disableKeyboardResize`, so the handle keeps advertising keyboard affordances it no longer has. Verified renders on this branch:

- corner handle → `role="button" tabindex="-1" aria-keyshortcuts="ArrowUp ArrowDown ArrowLeft ArrowRight"` — a non-focusable button promising arrow keys that now do nothing.
- edge handle → `role="separator" tabindex="-1" aria-valuenow="300"`.

Per ARIA, a separator is a widget only while it is focusable; a non-focusable one is a static structural separator and does not support `aria-valuenow` / `aria-valuemin` / `aria-valuemax` / `aria-orientation`. axe-core flags that combination.

When `disableKeyboardResize` is set and `tabIndex` is not overridden, drop `aria-keyshortcuts` on corner handles and the separator value attributes on edge handles.

packages/components/src/components/Resizable/hooks/useResizableState.ts
```diff
@@ -72,16 +72,40 @@ export const useResizableState = (
...
   const [managedSize, setManagedSize] = useControlledState<
-    ResizableSize | null,
-    ResizableSize
-  >(controlledSize, defaultSize ?? null, onResize);
+    ResizableSizeConstraints | null,
+    ResizableSizeConstraints
+  >(controlledSize, defaultSize ?? null);
+
+  const managedSizeRef = useRef(managedSize);
+  managedSizeRef.current = managedSize;
```
lskramarov: **correctness** — ref written during render, and it is avoidable.

Assigning `managedSizeRef.current` in the render body is the impurity React explicitly warns about: a render that React starts and discards (StrictMode double-render, an interrupted transition, an Offscreen/prerendered subtree) still writes the ref, so `applySize` can merge against a `managedSize` that was never committed and write a stale value into the axis it is preserving.

`useControlledState`’s setter already supports a functional updater fed from its own `valueRef`, which is reset in a `useInsertionEffect` after every render and so tracks the controlled prop correctly. That removes both the ref and its render-body assignment:

```tsx
setManagedSize((prev) => ({
  ...prev,
  ...(x !== 0 && { width: nextSize.width }),
  ...(y !== 0 && { height: nextSize.height }),
}));
```

packages/components/src/components/Resizable/hooks/useResizableHandle.ts
```diff
@@ -55,6 +61,14 @@ export const useResizableHandle = (
     onMoveEnd,
   });
 
+  // `useMove` drives both pointer and arrow keys, so dropping its key handler
+  // leaves dragging intact.
+  const interactionProps = { ...moveProps };
+
+  if (disableKeyboardResize) {
+    delete interactionProps.onKeyDown;
```
lskramarov: **altitude** — deleting a key out of React Aria’s memoized `moveProps` hard-codes an assumption about `useMove`’s internals.

AGENTS.md notes React Aria packages are pinned to exact versions and bumped together. The day `useMove` moves any part of its keyboard handling to `onKeyUp`, splits it, or renames the prop, this `delete` silently stops disabling anything — no type error, and no test failure either, since the assertion only checks that the size did not change. The `{ ...moveProps }` copy is also allocated on every render even when the flag is off.

An explicit override survives a rename and drops the unconditional copy (`mergeProps` resolves `undefined` to the other side, so a consumer-supplied `onKeyDown` is still preserved):

```suggestion
  // `useMove` drives both pointer and arrow keys, so dropping its key handler
  // leaves dragging intact.
  const interactionProps = disableKeyboardResize
    ? { ...moveProps, onKeyDown: undefined }
    : moveProps;
```

Gating at the callback boundary instead (`event.pointerType === 'keyboard'` in `onMoveStart` / `onMove`) would express the intent in the layer this package actually owns.

packages/components/src/components/Resizable/Resizable.mdx
```diff
@@ -54,6 +54,23 @@ The measured size is fixed in pixels only after the first resize interaction.
...
+
+```tsx
+<Resizable defaultSize={{ width: 320 }} minSize={{ width: 240 }}>
+  <Content />
+  <Resizable.Handle direction={[1, 0]} />
+</Resizable>
+```
+
+An axis becomes managed as soon as it's given in `size` / `defaultSize`, or resized by a handle.
+The `onResize`, `onResizeStart` and `onResizeEnd` handlers always report both dimensions,
```
lskramarov: **correctness** — “measured where an axis isn’t managed” overstates what the handlers deliver.

`resize` reads the unmanaged axis out of `startSizeRef`, so every `onResize` / `onResizeEnd` during a drag reports the value captured when the drag started, not a live measurement.

Verified: `defaultSize={{ width: 300 }}`, dragging the `[1, 0]` handle; mid-drag the container grows and the observed height goes 200 → 900. The final `onResize` still reports `{ width: 320, height: 200 }`.

That is the normal case for the side-panel example this section is written around — the panel’s height tracks its container, so any layout shift during a drag makes the reported height wrong. Either say “sampled when the resize starts”, or read the unmanaged axis from the live observed size in `applySize`.

packages/components/src/components/Resizable/Resizable.test.tsx
```diff
@@ -280,7 +280,100 @@ describe('Resizable', () => {
...
+      </Resizable>
+    );
+
+    fireEvent.keyDown(getHandle(), { key: 'ArrowDown' });
+
+    expect(getRoot()).toHaveStyle({ width: '300px', height: '201px' });
+  });
+
+  it('disables arrow key resizing without breaking dragging', () => {
+    renderResizable([1, 0], { defaultSize: { width: 300 } });
```
lskramarov: **test-quality** — leftover render; its result is never used.

This call (and the identical one on line 360) mounts a whole second `Resizable` before the actual subject is rendered below, so two elements end up carrying `data-testid="resizable"`. Verified: `getAllByTestId('resizable').length === 2`.

That is why these two tests had to reach for `handle.parentElement` instead of the file’s own `getRoot()` — the helper now throws `Found multiple elements` here. Anyone extending either test with the standard helpers hits a failure that has nothing to do with what is being tested.

```suggestion
```

packages/components/src/components/Resizable/Resizable.test.tsx
```diff
@@ -280,7 +280,100 @@ describe('Resizable', () => {
 
     fireEvent.keyDown(getHandle(), { key: 'ArrowRight' });
 
-    expect(getRoot()).toHaveStyle({ width: '261px', height: '140px' });
+    // Only the resized axis becomes managed, the other one is left to CSS.
+    expect(getRoot()).toHaveStyle({ width: '261px' });
+    expect(getRoot().style.height).toBe('');
+  });
+
+  it('manages a single dimension when the size defines one axis', () => {
```
lskramarov, line 261: **test-coverage** — every new single-axis test is uncontrolled.

Nothing here exercises `size` + `onResize` with one axis omitted, even though the suite already has a controlled test to model it on (line 247). The controlled round-trip is the case that is actually broken — echoing `onResize` back into `size` re-freezes the omitted axis after one drag (see the comment on `useResizableState.ts`).

A test asserting `getRoot().style.height === ''` after a controlled `[1, 0]` drag would have caught it, and is the guard that keeps it fixed.

packages/components/src/components/Resizable/Resizable.mdx
```diff
@@ -54,6 +54,23 @@ The measured size is fixed in pixels only after the first resize interaction.
 
 <Story of={Stories.IntrinsicSize} />
 
+### One dimension
+
+`size` and `defaultSize` accept a single dimension. The omitted one keeps its CSS size and is
+never written to the element, which suits elements sized by their layout in the other axis —
+a side panel stretched to the full height of its container, for example:
+
+```tsx
```
lskramarov: **conventions** — new features documented without a story.

AGENTS.md fixes the MDX shape as “`## Props`, then one section per feature with `<Story of={Stories.Feature} />`”. This section uses a hand-written ```` ```tsx ```` fence instead, and `disableKeyboardResize` gets two lines of prose under `## Keyboard interaction` with no story at all — `Resizable.stories.tsx` is untouched by this PR.

So two new public API surfaces ship with nothing runnable in Storybook: readers cannot see single-axis sizing behave, and the snippet is neither type-checked nor rendered, so it can drift from the component.

While adding the stories: AGENTS.md also asks that an updated component carry `status:updated` with a fresh `date:YYYY-MM-DD` — `meta.tags` is still `['status:new', 'date:2026-07-16']`.

packages/components/src/components/Resizable/Resizable.mdx
```diff
@@ -54,6 +54,23 @@ The measured size is fixed in pixels only after the first resize interaction.
 
 <Story of={Stories.IntrinsicSize} />
 
+### One dimension
```
lskramarov, line 57: **docs** — the `### Intrinsic size` text just above is now stale.

Line 53 still reads “The measured size is fixed in pixels only after the first resize interaction”, which describes the pre-PR behavior. This PR changes exactly that, and its own test edit proves it: the intrinsic-size assertion went from `{ width: '261px', height: '140px' }` to width plus `style.height === ''`.

A reader following `### Intrinsic size` still expects both axes to freeze after one drag and gets the opposite, with the contradicting rule buried in the next section. Suggest “the resized axis is fixed in pixels after the first resize interaction”.

packages/components/src/components/Resizable/hooks/utils.ts
```diff
@@ -53,10 +53,35 @@ export const clampResizableSize = (
...
+  size: ResizableSizeConstraints | undefined,
   bounds: ResizableBounds
-) => (size ? clampResizableSize(size, bounds) : undefined);
+): ResizableSizeConstraints | undefined => {
+  if (!size) return undefined;
+
+  const { width, height } = size;
+
+  return {
+    ...(width !== undefined && {
```
lskramarov, line 64: **reuse** — the per-axis clamp rule now exists four times in this file.

This inlines the same `clamp(isFiniteNumber(v) ? v : min, min, max)` expression that `clampResizableSize` above already contains, once per axis in each.

The non-finite fallback (`NaN` / `Infinity` → the minimum) is exactly the kind of rule that gets changed in one copy and missed in the others, and the `normalizes negative and non-finite sizes to the minimum` test only covers the `normalizeResizableSize` path — a divergence in `clampResizableSize` would go unnoticed.

Extracting the shared helper collapses both functions:

```tsx
const clampAxis = (value: number | undefined, min: number, max: number) =>
  clamp(isFiniteNumber(value) ? value : min, min, max);
```

packages/components/src/components/Resizable/hooks/useResizable.ts
```diff
@@ -56,18 +58,27 @@ export const useResizable = <T extends HTMLElement = HTMLElement>(
...
       const rect = targetRef.current?.getBoundingClientRect();
 
-      startResize(
-        direction,
-        rect ? { width: rect.width, height: rect.height } : currentSize
-      );
+      // A managed axis is the source of truth, the measured one may lag behind
+      // it while the element animates to the size that was set last.
+      startResize(direction, {
+        width: managedSize?.width ?? rect?.width ?? currentSize.width,
```
lskramarov: **correctness** — the snapshot now mixes two measurement systems for the same axis.

Managed axes come from state, unmanaged ones from `getBoundingClientRect()`, while `currentSize` — and therefore `aria-valuenow` and all clamping — comes from `useElementSize`’s untransformed border-box.

With a CSS `transform: scale()` on the Resizable or any ancestor, `rect.width` / `rect.height` is the scaled visual size while the ResizeObserver border-box is the unscaled layout size. Under `scale(0.5)` a 400px-tall unmanaged panel snapshots as 200, so the first drag on a `[0, 1]` handle jumps the height to roughly half its real value, and `onResizeStart` reports a figure that does not match the `aria-valuenow` a screen reader just announced.

Reading the unmanaged axis from the already-observed `width` / `height` keeps one source of truth — and makes the `currentSize.width` / `currentSize.height` tail of the fallback chain unreachable, so it can go too.

packages/components/src/components/Resizable/hooks/useResizableState.ts
KamilEmeleev: Kept the handlers reporting both dimensions: reporting only the managed axes would make `width` / `height` optional in every callback, 2D included. To control a single axis, take only that axis from the handler — `onResize={(size) => setWidth(size.width)}` — as shown in the new `OneDimension` story. A test covers the controlled round-trip.

packages/components/src/components/Resizable/Resizable.test.tsx
KamilEmeleev: Added `keeps the omitted axis unmanaged in controlled state`: after a controlled `[1, 0]` drag, `style.height` stays empty.

packages/components/src/components/Resizable/hooks/useResizableHandle.ts
```diff
@@ -55,6 +61,14 @@ export const useResizableHandle = (
...
+  }
```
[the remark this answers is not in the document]
KamilEmeleev: Fixed: drag-only corner handles no longer set `aria-keyshortcuts`, covered by a test.

packages/components/src/components/Resizable/hooks/useResizableHandle.ts
KamilEmeleev: Dropped `aria-keyshortcuts` on drag-only corner handles. The separator values stay: with `tabindex="-1"` the handle is still focusable, just not tabbable, and a focusable separator must have `aria-valuenow`.

packages/components/src/components/Resizable/hooks/useResizable.ts
KamilEmeleev: Done: a resize now starts from `currentSize` — managed axes from state, the others from the observed border-box — so it matches `aria-valuenow` and ignores transforms. The `getBoundingClientRect` mock is gone from the tests too.

packages/components/src/components/Resizable/Resizable.mdx
KamilEmeleev: Reworded: an axis that isn't managed is measured once, when the resize starts. Reading it live would add plumbing for a value single-axis consumers don't use.

packages/components/src/components/Resizable/Resizable.mdx
KamilEmeleev: Reworded: the first resize interaction fixes the resized axes in pixels, the others keep their CSS size.

packages/components/src/components/Resizable/Resizable.mdx
KamilEmeleev: Replaced the snippet with a `OneDimension` story — a controlled, width-only panel stretched to the container height — and switched the tags to `status:updated` / `date:2026-09-11`. There's no story for `disableKeyboardResize`: it only removes keyboard behavior, so there's nothing to show.

packages/components/src/components/Resizable/hooks/useResizable.ts
KamilEmeleev: Done as suggested, which also removes the second clamp.

packages/components/src/components/Resizable/hooks/useResizableState.ts
KamilEmeleev: Done: `setManagedSize((prevSize) => …)`, the ref is gone.

packages/components/src/components/Resizable/hooks/useResizableHandle.ts
KamilEmeleev: Took your version. Gating in `onMoveStart` / `onMove` wouldn't be enough: `useMove` calls `preventDefault()` and `stopPropagation()` on arrow keys, so they'd still be swallowed.

packages/components/src/components/Resizable/hooks/utils.ts
KamilEmeleev: Done, both functions use `clampAxis` now.

packages/components/src/components/Resizable/Resizable.test.tsx
KamilEmeleev: Removed the leftover renders in both tests; they use `getRoot()` / `getHandle()` now.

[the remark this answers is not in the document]
KamilEmeleev: Removed, along with the one on line 360.

lskramarov (approved):

PR #484, opened 2026-09-11
docs(components): add highlighting matches examples (DS-5294)

KamilEmeleev:
## Summary by CodeRabbit

* **New Features**
  * Added examples demonstrating matching-text highlighting across autocomplete, dropdown menu, select, tag autocomplete, and tree select components.
  * Highlighted search terms now update as users type in the documented examples.

* **Documentation**
  * Added guidance for using `Highlight`, providing plain-text labels with `textValue`, and updating results as queries change.
  * Clarified filtering behavior and requirements for refreshed option data.
  * Replaced dependency-focused examples with highlighting examples where applicable.

### DIFF
--- a/packages/components/src/components/Autocomplete/Autocomplete.mdx
+++ b/packages/components/src/components/Autocomplete/Autocomplete.mdx
@@ -96,9 +96,9 @@ The following example uses the `defaultFilter` prop to filter the list of option
 
 ### Highlighting matches
 
-Use [Highlight](/docs/components-highlight--docs) to mark the part of an option that matched the query.
-Control `inputValue` so the query is available for both filtering and highlighting, and set `textValue`
-on the item so the collection keeps a plain-text label for typeahead and accessibility.
+Put [Highlight](/docs/components-highlight--docs) into `Autocomplete.ItemText` to mark the matched
+part of an option. Build the options in a `useMemo` keyed by the query — otherwise the highlight
+does not update.
 
 <Story of={Stories.HighlightingMatches} />
 
--- a/packages/components/src/components/Autocomplete/Autocomplete.stories.tsx
+++ b/packages/components/src/components/Autocomplete/Autocomplete.stories.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useState } from 'react';
+import { useCallback, useMemo, useState } from 'react';
 
 import { isString, useDebounceCallback } from '@koobiq/react-core';
 import {
@@ -276,31 +276,35 @@ export const CustomFiltering: Story = {
 
 export const HighlightingMatches: Story = {
   render: function Render() {
-    const items = [
-      { key: 'tls', name: 'TLS' },
-      { key: 'ssh', name: 'SSH' },
-      { key: 'pgp', name: 'PGP' },
-      { key: 'ipsec', name: 'IPSec' },
-      { key: 'kerberos', name: 'Kerberos' },
-    ];
-
     const { contains } = useFilter({ sensitivity: 'base' });
 
     const [inputValue, setInputValue] = useState('');
 
-    const filtered = items.filter((item) => contains(item.name, inputValue));
+    const items = useMemo(
+      () =>
+        [
+          { key: 'tls', name: 'TLS' },
+          { key: 'ssh', name: 'SSH' },
+          { key: 'pgp', name: 'PGP' },
+          { key: 'ipsec', name: 'IPSec' },
+          { key: 'kerberos', name: 'Kerberos' },
+        ].filter((item) => contains(item.name, inputValue)),
+      [contains, inputValue]
+    );
 
     return (
       <Autocomplete
+        items={items}
         label="Protocol"
-        items={filtered}
-        placeholder="Search a protocol"
         inputValue={inputValue}
         onInputChange={setInputValue}
+        placeholder="Search a protocol"
       >
         {(item) => (
           <Autocomplete.Item key={item.key} textValue={item.name}>
-            <Highlight text={item.name} query={inputValue} />
+            <Autocomplete.ItemText>
+              <Highlight text={item.name} query={inputValue} />
+            </Autocomplete.ItemText>
           </Autocomplete.Item>
         )}
       </Autocomplete>
--- a/packages/components/src/components/DropdownMenu/DropdownMenu.stories.tsx
+++ b/packages/components/src/components/DropdownMenu/DropdownMenu.stories.tsx
@@ -1,5 +1,5 @@
 import type { CSSProperties } from 'react';
-import { useState } from 'react';
+import { useMemo, useState } from 'react';
 
 import { useBoolean } from '@koobiq/react-core';
 import {
@@ -26,6 +26,7 @@ import { Button } from '../Button';
 import { Divider } from '../Divider';
 import { EmptyState } from '../EmptyState';
 import { FlexBox } from '../FlexBox';
+import { Highlight } from '../Highlight';
 import { spacing } from '../layout';
 import { SearchInput } from '../SearchInput';
 import { SelectNext as Select } from '../SelectNext';
@@ -619,6 +620,54 @@ export const SearchEmpty: Story = {
       </DropdownMenu.Popover>
     </DropdownMenu>
   ),
+};
+
+export const HighlightingMatches: Story = {
+  render: function Render(args) {
+    const [inputValue, setInputValue] = useState('');
+
+    const items = useMemo(
+      () => [
+        { id: 'amsterdam', name: 'Amsterdam' },
+        { id: 'belgrade', name: 'Belgrade' },
+        { id: 'berlin', name: 'Berlin' },
+        { id: 'bratislava', name: 'Bratislava' },
+        { id: 'brussels', name: 'Brussels' },
+        { id: 'bucharest', name: 'Bucharest' },
+        { id: 'budapest', name: 'Budapest' },
+        { id: 'copenhagen', name: 'Copenhagen' },
+      ],
+      []
+    );
+
+    return (
+      <DropdownMenu {...args}>
+        <Button>Cities</Button>
+        <DropdownMenu.Popover>
+          <DropdownMenu.Autocomplete
+            inputValue={inputValue}
+            onInputChange={setInputValue}
+          >
+            <SearchInput />
+            <Divider disablePaddings />
+            <DropdownMenu.Content
+              items={items}
+              dependencies={[inputValue]}
+              onAction={(key) => alert(key)}
+            >
+              {(item) => (
+                <DropdownMenu.Item id={item.id} textValue={item.name}>
+                  <DropdownMenu.ItemText>
+                    <Highlight text={item.name} query={inputValue} />
+                  </DropdownMenu.ItemText>
+                </DropdownMenu.Item>
+              )}
+            </DropdownMenu.Content>
+          </DropdownMenu.Autocomplete>
+        </DropdownMenu.Popover>
+      </DropdownMenu>
+    );
+  },
 };
 
 export const DropdownFooter: Story = {
--- a/packages/components/src/components/SelectNext/Select.mdx
+++ b/packages/components/src/components/SelectNext/Select.mdx
@@ -186,22 +186,6 @@ with `defaultInputValue`, and the `onInputChange` callback is called whenever th
 
 <Story of={Stories.Searchable} />
 
-### Dependencies
-
-The collection caches a rendered item by the identity of its data object, so an item is not
-re-rendered when a value used inside the render function changes. List such values in
-`dependencies` to invalidate that cache — for example the search query when the options
-highlight it.
-
-Keep the length of the array the same between renders, otherwise the items are not
-re-rendered. To depend on a list of values, wrap it: `dependencies={[filters]}`.
-
-The same applies to `Select.Section`: it inherits `dependencies` from the Select, so the
-prop only has to be set once, on the root. A section written out in JSX can also take its
-own `dependencies`, for values only its items use.
-
-<Story of={Stories.Dependencies} />
-
 ### Minimum options for search
 
 Short lists don't need a search input. The `minOptionsThreshold` prop shows the search only when
@@ -209,6 +193,13 @@ there are at least that many options. It works together with `isSearchable`. The
 so the search is always shown.
 
 <Story of={Stories.SearchableMinOptionsThreshold} />
+
+### Highlighting matches
+
+Put [Highlight](/docs/components-highlight--docs) into `Select.ItemText` to mark the matched part of
+an option. Pass the query in `dependencies` — otherwise the highlight does not update.
+
+<Story of={Stories.HighlightingMatches} />
 
 ### Server search
 
--- a/packages/components/src/components/SelectNext/Select.stories.tsx
+++ b/packages/components/src/components/SelectNext/Select.stories.tsx
@@ -1,4 +1,4 @@
-import { useState, useCallback } from 'react';
+import { useState, useCallback, useMemo } from 'react';
 
 import { useBoolean } from '@koobiq/react-core';
 import {
@@ -544,30 +544,6 @@ export const Searchable: Story = {
   },
 };
 
-export const Dependencies: Story = {
-  render: function Render() {
-    const [inputValue, setInputValue] = useState('');
-
-    return (
-      <Select
-        items={options}
-        label="Attack type"
-        dependencies={[inputValue]}
-        onInputChange={setInputValue}
-        style={{ inlineSize: 200 }}
-        placeholder="Select an option"
-        isSearchable
-      >
-        {(item) => (
-          <Select.Item id={item.id} textValue={item.name}>
-            <Highlight text={item.name} query={inputValue} />
-          </Select.Item>
-        )}
-      </Select>
-    );
-  },
-};
-
 export const SearchableMinOptionsThreshold: Story = {
   render: function Render() {
     return (
@@ -593,6 +569,45 @@ export const SearchableMinOptionsThreshold: Story = {
           {(item) => <Select.Item id={item.id}>{item.name}</Select.Item>}
         </Select>
       </FlexBox>
+    );
+  },
+};
+
+export const HighlightingMatches: Story = {
+  render: function Render() {
+    const [inputValue, setInputValue] = useState('');
+
+    const items = useMemo(
+      () => [
+        { id: 1, name: 'Bruteforce' },
+        { id: 2, name: 'Complex Attack' },
+        { id: 3, name: 'DDoS' },
+        { id: 4, name: 'HIPS Alert' },
+        { id: 5, name: 'Network Attack' },
+        { id: 6, name: 'Potential Attack' },
+      ],
+      []
+    );
+
+    return (
+      <Select
+        items={items}
+        label="Attack type"
+        inputValue={inputValue}
+        dependencies={[inputValue]}
+        style={{ inlineSize: 200 }}
+        onInputChange={setInputValue}
+        placeholder="Select an option"
+        isSearchable
+      >
+        {(item) => (
+          <Select.Item id={item.id} textValue={item.name}>
+            <Select.ItemText>
+              <Highlight text={item.name} query={inputValue} />
+            </Select.ItemText>
+          </Select.Item>
+        )}
+      </Select>
     );
   },
 };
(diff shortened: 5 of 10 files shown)


lskramarov:
@rmnturov нам точно нужны примеры с поиском но без выделения текста в доке ?

packages/components/src/components/SelectNext/Select.stories.tsx
```diff
@@ -597,6 +573,45 @@ export const SearchableMinOptionsThreshold: Story = {
...
+      ],
+      []
+    );
+
+    return (
+      <Select
+        items={items}
+        label="Attack type"
+        inputValue={inputValue}
+        dependencies={[inputValue]}
```
lskramarov, line 597: `inputValue={inputValue}` makes the search input controlled, and that silently disables Select's own reset-on-close: the effect in `SelectInner` bails out when `inputValueProp !== undefined`, so the query is never cleared after the dropdown closes.

Verified in Storybook: open, type `att`, close, reopen → the search box still holds `att` and the list is still filtered to 3 options. The uncontrolled `Searchable` story reopens with an empty query and all 11 options.

Nothing here needs the controlled value — `onInputChange` already feeds both `Highlight` and `dependencies`, and the prose above the story only mentions `dependencies`.

```suggestion
        dependencies={[inputValue]}
```

packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
```diff
@@ -601,27 +601,52 @@ export const Searchable: Story = {
...
+      []
+    );
+
     return (
       <TreeSelect
-        items={items}
+        items={files}
         label="Project files"
+        inputValue={inputValue}
         dependencies={[inputValue]}
```
lskramarov, line 635: Same problem as in the Select example: making the query controlled disables TreeSelect's reset-on-close, which is guarded by `inputValueProp !== undefined` in `TreeSelectInner`.

Verified in Storybook: open, type `app`, dismiss by clicking outside, reopen → the search box still holds `app`, the tree is still filtered to `app` / `app.js`, and the old matches are still highlighted.

`onInputChange` alone is enough for `Highlight` and `dependencies`.

```suggestion
        dependencies={[inputValue]}
```

packages/components/src/components/DropdownMenu/DropdownMenu.stories.tsx
```diff
@@ -621,6 +622,54 @@ export const SearchEmpty: Story = {
...
+      ],
+      []
+    );
+
+    return (
+      <DropdownMenu {...args}>
+        <Button>Cities</Button>
+        <DropdownMenu.Popover>
+          <DropdownMenu.Autocomplete
+            inputValue={inputValue}
```
lskramarov, line 648: Here the controlled `inputValue` is genuinely needed (it feeds `Highlight` and `dependencies`), but nothing ever resets it, so the query outlives the popover. Uncontrolled, RAC's `Autocomplete` state is thrown away when `DropdownMenu.Popover` unmounts.

Verified in Storybook: open, type `bu`, dismiss by clicking outside, reopen → the search box still holds `bu` and the menu is still filtered to Bucharest / Budapest. The existing uncontrolled `Search` story reopens empty with all 22 items.

Clearing the query when the menu closes keeps the example faithful to the default behaviour — e.g. `<DropdownMenu {...args} onOpenChange={(isOpen) => { if (!isOpen) setInputValue(''); }}>`.

packages/components/src/components/Autocomplete/Autocomplete.mdx
```diff
@@ -96,9 +96,9 @@ The following example uses the `defaultFilter` prop to filter the list of option
 
 ### Highlighting matches
 
-Use [Highlight](/docs/components-highlight--docs) to mark the part of an option that matched the query.
-Control `inputValue` so the query is available for both filtering and highlighting, and set `textValue`
-on the item so the collection keeps a plain-text label for typeahead and accessibility.
+Put [Highlight](/docs/components-highlight--docs) into `Autocomplete.ItemText` to mark the matched
+part of an option. Build the options in a `useMemo` keyed by the query — otherwise the highlight
```
lskramarov, line 100: This rule is stated in terms of the wrong mechanism, and a reader following it literally will still get a stale highlight.

`CollectionBuilder` caches a rendered node by the **identity of the item object** (`this.cache.get(partialNode.value)` in `@react-stately/collections`), and `Autocomplete` has no `dependencies` escape hatch. The story works because the object literals are *created inside* the `useMemo`, so every keystroke produces brand-new objects — not because the memo is keyed by the query.

Written the way this sentence reads, it silently breaks:

```tsx
const PROTOCOLS = [{ key: 'tls', name: 'TLS' }, /* … */];

// useMemo keyed by the query, yet the highlight never updates:
// the surviving objects are cache hits.
const items = useMemo(
  () => PROTOCOLS.filter((item) => contains(item.name, inputValue)),
  [contains, inputValue]
);
```

Worth saying what actually matters — the item objects have to be new when the query changes (which also means the pre-PR version of this story, which rebuilt the array on every render, was not broken either). The same sentence appears in `TagAutocomplete.mdx`.

packages/components/src/components/TagAutocomplete/TagAutocomplete.mdx
```diff
@@ -79,6 +79,14 @@ Selecting a suggestion calls `onAdd(values, context)` with
 `context.source === 'suggestion'` and exposes the original item as
 `context.suggestion`.
 
+### Highlighting matches
+
+Put [Highlight](/docs/components-highlight--docs) into `TagAutocomplete.ListItem` to mark the
+matched part of a suggestion. Build the suggestions in a `useMemo` keyed by the query — otherwise
```
lskramarov, line 85: Same wording problem as in `Autocomplete.mdx`: it is not the `useMemo` that makes the highlight update, it is the fact that the suggestion objects are re-created inside it. `useTagAutocompleteState` → `useListState` → `CollectionBuilder` caches each rendered node by item-object identity, so `useMemo(() => SUGGESTIONS.filter(…), [inputValue])` over a stable module-level array is keyed by the query and still renders a stale highlight.

packages/components/src/components/Autocomplete/Autocomplete.mdx
```diff
@@ -96,9 +96,9 @@ The following example uses the `defaultFilter` prop to filter the list of option
...
+does not update.
```
lskramarov, line 101: The rewrite drops the `textValue` half of the old paragraph ("set `textValue` on the item so the collection keeps a plain-text label for typeahead and accessibility"), and none of the four new sections in this PR mention it either.

It is load-bearing in every one of the five examples: once the item's children are a `<Highlight>` element instead of a string, `textValue` is the only source for filtering, typeahead and the option's accessible name. A reader who copies the snippet without it gets a list that stops filtering and an option with no accessible label — a silent failure, not a type error.

packages/components/src/components/SelectNext/Select.mdx
```diff
@@ -210,6 +194,13 @@ so the search is always shown.
 
 <Story of={Stories.SearchableMinOptionsThreshold} />
 
+### Highlighting matches
```
lskramarov, line 197: This replaces, rather than complements, the `### Dependencies` section — and `dependencies` shipped only one commit earlier (2b51a4ac, #475). Two things that had no other home are now gone from the docs:

- `Select.Section` inherits `dependencies` from the root, so the prop only has to be set once — and a JSX section can carry its own `dependencies`. Nothing in the generated Props table says this.
- The `Dependencies` story was the only worked example of the prop.

Highlighting is the motivating case, not the whole feature (`dependencies` matters for any value read inside the render function). Keeping both sections — `Dependencies` for the mechanism, `Highlighting matches` for the recipe — seems better than folding the mechanism into a single half-sentence.

packages/components/src/components/TreeSelect/TreeSelect.mdx
```diff
@@ -154,20 +154,12 @@ and matches text anywhere in an item. To change how items match, pass your own
 
 <Story of={Stories.Searchable} />
 
-### Dependencies
+### Highlighting matches
```
lskramarov, line 157: Same as in `Select.mdx`: the `### Dependencies` section is deleted outright, one commit after the prop was added (2b51a4ac, #475). The note it carried — "Nested `Collection` elements inherit `dependencies` from the TreeSelect, so the prop only has to be set once, on the root" — is not stated anywhere else, and it is exactly the question a reader of the new story will have, since the story renders `<Collection items={item.children}>` without `dependencies`.

packages/components/src/components/TreeSelect/TreeSelect.mdx
```diff
@@ -154,20 +154,12 @@ and matches text anywhere in an item. To change how items match, pass your own
...
 
-The collection caches a rendered item by the identity of its data object, so an item is not
-re-rendered when a value used inside the render function changes. List such values in
-`dependencies` to invalidate that cache — for example the search query when the items
-highlight it.
+Put [Highlight](/docs/components-highlight--docs) into `Tree.ItemContentText` to mark the matched
```
lskramarov, line 159: `Tree.ItemContentText` is not a slot of `TreeSelect` — the compound only assigns `Item`, `ItemContent`, `LoadMoreItem` and `Tag` (`TreeSelect.tsx`), so a consumer following this page has to import `Tree` as well, on a page that otherwise says `TreeSelect.Item` / `TreeSelect.ItemContent`.

`Select` handles the equivalent case by exposing `Select.ItemText`. Adding `ItemContentText` (and `ItemContentAddon`, which is in the same position) to the `TreeSelect` compound would let this read `TreeSelect.ItemContentText` and let the slot appear in `meta.subcomponents`, which `AGENTS.md` asks for ("Add every slot of a compound component to `meta.subcomponents`").

packages/components/src/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -276,31 +276,35 @@ export const CustomFiltering: Story = {
...
+          { key: 'pgp', name: 'PGP' },
+          { key: 'ipsec', name: 'IPSec' },
+          { key: 'kerberos', name: 'Kerberos' },
+        ].filter((item) => contains(item.name, inputValue)),
+      [contains, inputValue]
+    );
 
     return (
       <Autocomplete
+        items={items}
```
lskramarov, line 297: Passing the *pre-filtered* list as `items` (rather than `defaultItems`) hands the collection over to the story, and `useComboBoxState` reacts by turning its own machinery off:

```js
// react-stately/combobox
let filteredCollection = useMemo(() => (
  // No default filter if items are controlled.
  props.items != null || !defaultFilter ? triggerState.collection : filterCollection(...)
), [...]);
```

Two consequences the MDX doesn't mention:

- `defaultFilter` (and the `contains` default documented in "Custom filtering") is dead on this example — filtering is entirely the story's job.
- "Show all options" on a manual open is gone. `showAllItems` falls back to `originalCollection`, which here *is* the filtered list, so opening the dropdown with the chevron while the input holds `ss` still shows only SSH instead of all five protocols.

Neither is avoidable with today's API — `Autocomplete` has no `dependencies` prop, so pre-filtering is the only way to get fresh item identities. Worth at least calling the trade-off out in the section, so readers don't copy it into a case where the chevron behaviour matters.

packages/components/src/components/TagAutocomplete/TagAutocomplete.stories.tsx
```diff
@@ -107,6 +108,73 @@ export const Base: Story = {
...
+        [
+          { id: 'react', name: 'React' },
+          { id: 'typescript', name: 'TypeScript' },
+          { id: 'storybook', name: 'Storybook' },
+          { id: 'vite', name: 'Vite' },
+          { id: 'vitest', name: 'Vitest' },
+          { id: 'playwright', name: 'Playwright' },
+        ].filter((item) => containsFilter(item.name, inputValue)),
+      [inputValue]
+    );
```
lskramarov, line 134: This is the only story in the file that filters by hand instead of passing `defaultFilter={containsFilter}` — and it re-implements what `useTagAutocompleteState`'s built-in `filter` already does. It also contradicts the "Suggestions" section two paragraphs above the new one, which tells the reader that "`defaultFilter(textValue, inputValue)` filters the popover list".

The manual filter isn't what makes the highlight work — fresh item identities are. Both can be had by just building the array inside `render` and letting the component filter:

```suggestion
    const suggestions = [
      { id: 'react', name: 'React' },
      { id: 'typescript', name: 'TypeScript' },
      { id: 'storybook', name: 'Storybook' },
      { id: 'vite', name: 'Vite' },
      { id: 'vitest', name: 'Vitest' },
      { id: 'playwright', name: 'Playwright' },
    ];
```

…with `defaultFilter={containsFilter}` back on the component (and `useMemo` dropped from the import).

packages/components/src/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -276,31 +276,35 @@ export const CustomFiltering: Story = {
...
+      () =>
+        [
+          { key: 'tls', name: 'TLS' },
+          { key: 'ssh', name: 'SSH' },
+          { key: 'pgp', name: 'PGP' },
+          { key: 'ipsec', name: 'IPSec' },
+          { key: 'kerberos', name: 'Kerberos' },
+        ].filter((item) => contains(item.name, inputValue)),
+      [contains, inputValue]
+    );
```
lskramarov, line 293: Stepping back from this one story: the PR ends up documenting **two** different answers to the same question.

- `Select`, `TreeSelect` and `DropdownMenu.Content` take `dependencies`, so their examples keep a stable `useMemo(…, [])` list and declare the query — the shape the docs already teach.
- `Autocomplete` and `TagAutocomplete` have no such prop, so their examples have to churn item identities on every keystroke. That workaround is more fragile (it breaks the moment someone hoists the source array — see the `.mdx` comment), it throws away memoization, and for `Autocomplete` it also forces `items` over `defaultItems`, which disables the component's own filtering.

The deeper fix is to give `Autocomplete` and `TagAutocomplete` a `dependencies` prop too — they build their collections through `useCollection`/`CollectionBuilder`, whose cache is exactly what `dependencies` invalidates — and then all five pages say the same thing. Reasonable to land these examples first and follow up, but worth a ticket rather than leaving the second idiom as documented guidance.

rmnturov (changes requested):
В Кубике в поисковые совпадения выделяются фоном. Это должно работать в выпадашках (селекты, автокомплиты и тп). Примеры должны это отражать, чтобы люди знали, что считается поведением по умолчанию. То есть ожидаю, что в компонентах с поиском в сторибуке совпадения в поиске будут выбелены фоном.

packages/components/src/components/SelectNext/Select.stories.tsx
KamilEmeleev: Fixed — dropped `inputValue`, `onInputChange` alone feeds `Highlight` and `dependencies`. Reopening now starts empty with all 6 options.

The controlled mode itself still skips the reset, so that part is DS-5576.

packages/components/src/components/TreeSelect/TreeSelect.stories.tsx
KamilEmeleev: Fixed, same as in Select. Reopening starts empty with the full tree and no stale highlight. Controlled mode → DS-5576.

packages/components/src/components/DropdownMenu/DropdownMenu.stories.tsx
KamilEmeleev: Fixed with your `onOpenChange` reset. Reopening shows all 8 cities and an empty search box.

The component doing this on its own is DS-5576 — once it lands, the manual reset comes out of this story.

packages/components/src/components/Autocomplete/Autocomplete.mdx
KamilEmeleev: You're right, rewrote it: the options have to be new objects when the query changes, and filtering a constant array reuses them.

packages/components/src/components/TagAutocomplete/TagAutocomplete.mdx
KamilEmeleev: Rewritten the same way.

packages/components/src/components/Autocomplete/Autocomplete.mdx
KamilEmeleev: Restored, and added to all five sections — it is load-bearing in each of them.

packages/components/src/components/SelectNext/Select.mdx
KamilEmeleev: We're keeping the section folded for now. Describing the caching mechanism once, in a shared place, rather than per component page is part of DS-5529 — which also adds `dependencies` to `Autocomplete` and `TagAutocomplete`.

packages/components/src/components/TreeSelect/TreeSelect.mdx
KamilEmeleev: Same answer as in Select.mdx — DS-5529.

packages/components/src/components/TreeSelect/TreeSelect.mdx
KamilEmeleev: Agreed. DS-5575: add `TreeSelect.ItemContentText` and `TreeSelect.ItemContentAddon` to the compound, with `meta.subcomponents` and an API report update. Out of scope for a docs PR since it changes the public API.

packages/components/src/components/Autocomplete/Autocomplete.stories.tsx
KamilEmeleev: Called out in the section now: `defaultFilter` is not called and the chevron opens the filtered list instead of all options. The workaround itself goes away with DS-5529.

packages/components/src/components/TagAutocomplete/TagAutocomplete.stories.tsx
KamilEmeleev: Fixed — took your suggestion: the array is built inside `render` and `defaultFilter={containsFilter}` is back.

packages/components/src/components/Autocomplete/Autocomplete.stories.tsx
KamilEmeleev: Agreed, filed as DS-5529: `dependencies` for `Autocomplete` and `TagAutocomplete`, then all five pages say the same thing. Landing the examples first.

packages/components/src/components/Autocomplete/Autocomplete.mdx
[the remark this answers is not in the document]
KamilEmeleev: Rewritten — the rule is now about item object identity, not `useMemo`.

packages/components/src/components/SelectNext/Select.mdx
```diff
@@ -210,6 +194,13 @@ so the search is always shown.
...
+
+Put [Highlight](/docs/components-highlight--docs) into `Select.ItemText` to mark the matched part of
+an option. Pass the query in `dependencies` — otherwise the highlight does not update.
+
+<Story of={Stories.HighlightingMatches} />
+
```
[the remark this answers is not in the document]
KamilEmeleev: Not restoring the section here — the `dependencies` docs are being reworked in DS-5529.

packages/components/src/components/TreeSelect/TreeSelect.mdx
```diff
@@ -154,20 +154,12 @@ and matches text anywhere in an item. To change how items match, pass your own
...
+part of an item. Pass the query in `dependencies` — otherwise the highlight does not update.
```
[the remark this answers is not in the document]
KamilEmeleev: Not restoring the section here — the `dependencies` docs are being reworked in DS-5529.

packages/components/src/components/SelectNext/Select.stories.tsx
```diff
@@ -1,4 +1,4 @@
-import { useState, useCallback } from 'react';
+import { useState, useCallback, useMemo } from 'react';
```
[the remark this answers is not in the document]
KamilEmeleev: Declined. The `'use client'` convention covers component files; none of the 72 story files has the directive.

KamilEmeleev:
> В Кубике в поисковые совпадения выделяются фоном. Это должно работать в выпадашках (селекты, автокомплиты и тп). Примеры должны это отражать, чтобы люди знали, что считается поведением по умолчанию. То есть ожидаю, что в компонентах с поиском в сторибуке совпадения в поиске будут выбелены фоном.

Согласен, подсветка фоном должна быть везде, где есть поиск.

Сейчас её включает пользователь: сам оборачивает текст пункта в `Highlight`. По умолчанию
подсветки нет.

Отдельная задача DS-5590 — включить её по умолчанию во всех выпадашках с поиском.
Руками останется только для своего рендера пункта, плюс проп, чтобы выключить.

Примеры в этом PR пока показывают ручной способ — другого сегодня нет. После задачи
они будут подсвечивать сами, а секции останутся про то, как поменять вид подсветки.

rmnturov (approved):

PR #485, opened 2026-09-14
feat(components): add DescriptionList component (DS-5248)

KamilEmeleev:
## Summary by CodeRabbit

- **New Features**
  - Added the `DescriptionList` component for semantic term-and-description pairs.
  - Supports horizontal, vertical, responsive, and adaptive layouts.
  - Added configurable columns, alignment, long-text wrapping, and dynamic collections.
  - Includes `Group`, `Term`, and `Description` subcomponents with accessible native list semantics.

- **Documentation**
  - Added usage guidance and examples covering layouts, collections, and composition with clamped content.

- **Tests**
  - Added coverage for rendering, responsive behavior, styling, accessibility semantics, collections, and ref forwarding.


packages/components/src/components/DescriptionList/DescriptionList.module.css
```diff
@@ -0,0 +1,34 @@
+.base {
```
lskramarov, line 1: **correctness** — `alignItems="start"` / `justifyItems="start"` are never actually applied.

`.base` sets neither `align-items` nor `justify-items`, and there are no `.alignStart` / `.justifyStart` modifiers — `DescriptionList.tsx:67-70` only emits a class for `center` and `end`.

So with the documented defaults (`types.ts:42`, `types.ts:47`, and what `data-align-items` / `data-justify-items` advertise) the `dl` computes `align-items: normal` / `justify-items: normal`, which for grid items behaves as **stretch**, not `start`. Every `dt`/`dd` then fills its whole track and row: a description with a background, border or hit area paints the full column width instead of shrinking to its text. Toggling `justifyItems` from `"center"` back to `"start"` does not restore start alignment.

The tests only assert the `data-*` attributes (`DescriptionList.test.tsx:118-127`), so nothing catches this. Either add explicit `start` rules, or — better — drive both properties from CSS custom properties the way `Grid` already does (see the separate comment on `DescriptionList.tsx:67`), which makes every enum value including `start` work by construction.

packages/components/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,122 @@
...
+function DescriptionListRender<T extends object = object>(
+  props: Omit<DescriptionListProps<T>, 'ref'>,
+  ref?: Ref<HTMLDListElement>
+) {
+  const { items, children, dependencies, ...other } = props;
+
+  return (
+    <CollectionBuilder
+      content={
+        <Collection items={items} dependencies={dependencies}>
```
lskramarov, line 91: **correctness** — children that are not `DescriptionList.Group` are silently dropped, with no dev warning.

`DescriptionList.Term` / `Description` are plain `forwardRef`s, not collection components, so React Aria's hidden collection document discards them. That means the most natural markup — the shape of a plain HTML `<dl>` — renders an empty list:

```tsx
<DescriptionList>
  <DescriptionList.Term>Status</DescriptionList.Term>
  <DescriptionList.Description>New</DescriptionList.Description>
</DescriptionList>
// => <dl></dl>
```

The PR's own test (`DescriptionList.test.tsx:235-246`) locks this in, and the rule exists only as MDX prose (`DescriptionList.mdx:46-47`). The repo already has the mechanism for surfacing exactly this class of mistake — `once.warn` guarded by `NODE_ENV`, as in `ClampedList.tsx:51-58`:

```tsx
if (process.env.NODE_ENV !== 'production') {
  const invalid = Children.toArray(children).some(
    (child) => isValidElement(child) && child.type !== DescriptionListGroup
  );

  if (invalid) {
    once.warn(
      'DescriptionList: only "DescriptionList.Group" children are supported; other children are ignored.'
    );
  }
}
```

packages/components/src/components/DescriptionList/components/DescriptionListGroup/DescriptionListGroup.tsx
```diff
@@ -0,0 +1,38 @@
...
+    props: DescriptionListGroupProps,
+    ref: ForwardedRef<HTMLDivElement>
+  ) {
+    const { className, children, ...other } = props;
+
+    // `id` is the key of the group in the collection, not a DOM id.
+    return (
+      <div
+        {...other}
+        id={undefined}
```
lskramarov, line 29: **correctness** — `id={undefined}` is dead code inside a collection, and harmful outside one.

`@react-aria/collections` `Document.setProps` destructures `{ value, textValue, id, ...props }` and assigns only the rest to `node.props`, so `props.id` is *already* `undefined` by the time this render function runs — for every group, static or dynamic. The test at `DescriptionList.test.tsx:248-259` passes whether or not this line exists.

On the other path it does have an effect, and it is the wrong one: `createLeafComponent` only throws `"… cannot be rendered outside a collection"` when the render function declares **3** parameters; this one declares 2, so `<DescriptionList.Group id="x">` used standalone renders fine — but silently without the id. A consumer can therefore never give a group a DOM id for `aria-labelledby` or an anchor link.

```suggestion
    return (
      <div
        {...other}
```

packages/components/src/components/DescriptionList/components/DescriptionListGroup/DescriptionListGroup.tsx
```diff
@@ -0,0 +1,38 @@
...
+        ref={ref}
+        className={clsx(s.base, className)}
+        data-slot="group"
+      >
+        {children}
+      </div>
+    );
+  }
+);
```
lskramarov, line 38: **conventions** — missing explicit `displayName`.

AGENTS.md (Component Anatomy): *"Always set `X.displayName`."*

`createLeafComponent` falls back to `Result.displayName = render.name`, so this slot registers as `DescriptionListGroup` while its siblings register as `DescriptionList.Term` (`DescriptionListTerm.tsx:29`) and `DescriptionList.Description` (`DescriptionListDescription.tsx:29`) — inconsistent in React DevTools and in Storybook's `subcomponents` table. The repo's other `createLeafComponent` slot sets it explicitly for exactly this reason (`FileUploadItem.tsx:327`).

```suggestion
);

(DescriptionListGroup as { displayName?: string }).displayName =
  'DescriptionList.Group';
```

packages/components/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,122 @@
...
+      style={style}
+      className={clsx(
+        s.base,
+        orientation === 'vertical' && s.vertical,
+        alignItems === 'center' && s.alignCenter,
+        alignItems === 'end' && s.alignEnd,
+        justifyItems === 'center' && s.justifyCenter,
+        justifyItems === 'end' && s.justifyEnd,
+        className
+      )}
```
lskramarov, line 72: **reuse** — the four alignment modifier classes re-implement what `Grid` already does with two CSS custom properties.

`Grid.tsx:43-47` writes `'--grid-align-items': alignItems` / `'--grid-justify-items': justifyItems` into the inline style and consumes each in a single declaration. Here the same job costs four CSS rules (`DescriptionList.module.css:20-33`) plus four conditional `clsx` arms, and every future alignment value costs another class + rule + branch.

It is also what causes the `start` bug flagged on `DescriptionList.module.css:1`: with variables, `start`, `center` and `end` all work without enumerating them.

`style` is already being constructed at lines 54-57 for `--description-list-columns`, so this costs nothing extra:

```tsx
const style = {
  ...styleProp,
  '--description-list-align-items': alignItems,
  '--description-list-justify-items': justifyItems,
  ...(columns !== undefined && { '--description-list-columns': columns }),
} as CSSProperties;
```

and in the CSS module, `align-items: var(--description-list-align-items);` / `justify-items: var(--description-list-justify-items);` on `.base`, leaving only `.vertical` as a modifier class.

packages/components/src/components/DescriptionList/components/DescriptionListDescription/DescriptionListDescription.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { forwardRef } from 'react';
+
+import { clsx } from '@koobiq/react-core';
+
+import type { DescriptionListDescriptionProps } from '../../types';
+
+import s from './DescriptionListDescription.module.css';
+
+/** DescriptionList.Description — a description of the term, e.g. a field value. */
+export const DescriptionListDescription = forwardRef<
```
lskramarov, line 12: **simplification** — `Term` and `Description` are duplicates.

This file and `DescriptionListTerm.tsx` are identical apart from the tag, the imported CSS module and the `data-slot` value. Their stylesheets differ by only two declarations: `DescriptionListTerm.module.css:3-11` vs `DescriptionListDescription.module.css:3-11` repeat `@mixin typography text-normal`, `min-inline-size: 0`, `margin: 0` and `overflow-wrap: break-word`, differing only in `grid-column` and `color`.

Since nothing tests the slot styles, a wrapping or typography fix applied to one and not the other diverges silently. One shared renderer parameterised by tag/slot/class, and one CSS module with `.term` / `.description` sharing a common base, removes both copies.

packages/components/src/components/DescriptionList/types.ts
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Vertical alignment of terms and descriptions.
+   * @default 'start'
+   */
+  alignItems?: DescriptionListPropAlign;
+  /**
+   * Horizontal alignment of terms and descriptions.
+   * @default 'start'
+   */
+  justifyItems?: DescriptionListPropAlign;
```
lskramarov, line 49: **conventions** — only `orientation` is responsive; `alignItems`, `justifyItems` and `columns` are not.

`Grid.tsx:33-38` resolves `cols`, `gap`, `colGap`, `rowGap`, `alignItems` **and** `justifyItems` through `getResponsiveValue`, so the breakpoint-aware API already exists one component over. A consumer who writes `orientation={{ xs: 'vertical', m: 'horizontal' }}` will reach for `columns={{ xs: '1fr', m: 'repeat(4, 1fr)' }}` next and get a type error.

This is a brand-new public surface already locked by `DescriptionList.api.md`, so it is cheapest to make consistent now. Widening these to `T | ResponsiveValue<T>` is a one-line change each plus a `getResponsiveValue` call, mirroring how `orientation` is already handled at `DescriptionList.tsx:51-52`.

packages/components/src/components/DescriptionList/components/DescriptionListGroup/DescriptionListGroup.module.css
```diff
@@ -0,0 +1,8 @@
+.base {
+  display: grid;
+  grid-column: 1 / -1;
+  grid-template-columns: subgrid;
+  row-gap: var(--description-list-content-gap);
```
lskramarov, line 5: **correctness** — slot styles read private custom properties that only the root declares, with no fallback.

`--description-list-content-gap` here, and `--description-list-description-column` in `DescriptionListDescription.module.css:6`, exist only on `.base` of `DescriptionList.module.css`. Outside a `DescriptionList` root they resolve to nothing, the declaration becomes invalid at computed-value time, and it silently drops to `row-gap: normal` / `grid-column: auto` — a group composed on its own collapses its term and description into one cell with no gap.

That path is reachable: `createLeafComponent` only throws "outside a collection" for 3-parameter render functions, and this one takes 2, so `DescriptionList.Group` renders happily standalone; `Term` / `Description` are ordinary `forwardRef`s usable anywhere.

```suggestion
  row-gap: var(--description-list-content-gap, 0);
```

(and the matching fallback in `DescriptionListDescription.module.css:6`).

packages/components/src/components/DescriptionList/DescriptionList.test.tsx
```diff
@@ -0,0 +1,334 @@
...
+            {(item) => (
+              <DescriptionList.Group>
+                <DescriptionList.Term>{item.term}</DescriptionList.Term>
+              </DescriptionList.Group>
+            )}
+          </DescriptionList>
+        )
+      ).toThrow('Could not determine key for item');
+
+      consoleError.mockRestore();
```
lskramarov, line 295: **test-coverage** — the `console.error` spy is restored only on the success path.

If the assertion at line 283 ever fails (say a React Aria bump changes the message), it throws and line 295 never runs. `tools/vitest/setupTests.ts:7-10` relies on `failOnConsole({ shouldFailOnError: true })` to turn stray console output into failures, and no `restoreMocks` / `clearMocks` is configured in `vite.config.mts` or `tools/vitest/config.ts` — so every later test in the file would run with `console.error` silenced, and the first failure would mask the rest.

```suggestion
      vi.restoreAllMocks();
```

or, keeping the local handle, `onTestFinished(() => consoleError.mockRestore())` right after the spy is created.

packages/components/src/components/DescriptionList/DescriptionList.mdx
```diff
@@ -0,0 +1,149 @@
...
+<DescriptionList>
+  <DescriptionList.Group>
+    <DescriptionList.Term />
+    <DescriptionList.Description />
+  </DescriptionList.Group>
+</DescriptionList>
+```
+
+- `DescriptionList` — the list (`<dl>`). It is a grid of two columns:
+  terms on the left, descriptions on the right.
```
lskramarov, line 45: **docs** — "a grid of two columns" contradicts the default.

`types.ts:37` documents `@default 'repeat(4, 1fr)'`, `DescriptionList.module.css:2` sets it, and the description spans `2 / -1` — so by default the term takes a quarter of the width and the description three quarters, not half and half. The `Columns` story labels the same value `columns="repeat(4, 1fr)" (default)`.

```suggestion
- `DescriptionList` — the list (`<dl>`). It is a grid: the term takes the
  first column, the description takes the rest (`repeat(4, 1fr)` by default).
```

packages/components/src/components/DescriptionList/DescriptionList.tsx
```diff
@@ -0,0 +1,122 @@
...
+  collection: AriaCollection<Node<T>>;
+};
+
+function DescriptionListInner<T extends object>(
+  props: DescriptionListInnerProps<T>
+) {
+  const {
+    listRef,
+    collection,
+    orientation: orientationProp = 'horizontal',
```
lskramarov, line 39: **simplification** — `orientation` is defaulted twice.

This destructuring default and the `?? 'horizontal'` at line 52 can never both matter: when the prop is omitted, `getResponsiveValue('horizontal', …)` returns `'horizontal'` and the `??` is dead; when a responsive object matches no breakpoint, this default is dead and the `??` does the work. Two defaults for one value means a future change has to be made in both places to take effect — `Grid.tsx:33-38` keeps a single resolution step.

```suggestion
    orientation: orientationProp,
```

artembelik (approved):

packages/components/src/components/DescriptionList/components/DescriptionListGroup/DescriptionListGroup.tsx
```diff
@@ -0,0 +1,38 @@
...
+import type { DescriptionListGroupProps } from '../../types';
+
+import s from './DescriptionListGroup.module.css';
+
+class DescriptionListGroupNode extends CollectionNode<unknown> {
+  static readonly type = 'item';
+}
+
+/** DescriptionList.Group — wraps a term and its description. */
+export const DescriptionListGroup = createLeafComponent(
```
artembelik, line 18: в angular такого компонента нет, он действительно нужен для реализации?

packages/components/src/components/DescriptionList/DescriptionList.module.css
KamilEmeleev: Fixed: alignment now uses CSS variables, so all values work. The default is `stretch`, as the list looked before. Also added `overflow-wrap: anywhere`, so long strings don't overflow.

packages/components/src/components/DescriptionList/DescriptionList.tsx
KamilEmeleev: Done, same as in `Grid`.

packages/components/src/components/DescriptionList/DescriptionList.tsx
KamilEmeleev: Added a dev warning: `Term` and `Description` warn when they are outside `Group`. Checking children by type would also warn on wrappers and fragments.

packages/components/src/components/DescriptionList/types.ts
KamilEmeleev: Done: `columns`, `alignItems` and `justifyItems` now accept values per breakpoint, like in `Grid`.

packages/components/src/components/DescriptionList/DescriptionList.mdx
KamilEmeleev: Fixed.

packages/components/src/components/DescriptionList/components/DescriptionListGroup/DescriptionListGroup.tsx
KamilEmeleev: Fixed.

packages/components/src/components/DescriptionList/DescriptionList.tsx
KamilEmeleev: Fixed.

tools/public_api_guard/components/DescriptionList.api.md
```diff
@@ -0,0 +1,75 @@
...
+
+// @public (undocumented)
+export type DescriptionListProps<T extends object = object> = ExtendableProps<DescriptionListBaseProps<T>, ComponentPropsWithRef<'dl'>>;
+
+// @public (undocumented)
+export type DescriptionListTermProps = ComponentPropsWithRef<'dt'> & DataAttributeProps;
+
+// Warnings were encountered during analysis:
+//
+// packages/components/dist/components/DescriptionList/types.d.ts:14:5 - (ae-forgotten-export) The symbol "ResponsiveValue" needs to be exported by the entry point index.d.ts
```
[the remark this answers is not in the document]
KamilEmeleev: The same warning is in Grid, FlexBox, Container and Form. We'll fix it for all of them in a separate task (DS-5543).

packages/components/src/components/DescriptionList/components/DescriptionListGroup/DescriptionListGroup.tsx
KamilEmeleev: Keeping it. `id` is the item key: React Aria items like `Breadcrumb` and `Tab` don't put it into the DOM either. Also, without this line type-check fails.

packages/components/src/components/DescriptionList/components/DescriptionListGroup/DescriptionListGroup.module.css
KamilEmeleev: Keeping it. Outside the list the gap is already 0, so a fallback changes nothing. And the parts are not meant to be used outside the list.

packages/components/src/components/DescriptionList/components/DescriptionListDescription/DescriptionListDescription.tsx
KamilEmeleev: Keeping them separate: they are small and render different elements with different styles.

packages/components/src/components/DescriptionList/DescriptionList.test.tsx
KamilEmeleev: No leak: `failOnConsole` restores `console.error` after each test.

packages/components/src/components/DescriptionList/DescriptionList.module.css
```diff
@@ -0,0 +1,34 @@
...
+  --description-list-columns: repeat(4, 1fr);
+  --description-list-description-column: 2 / -1;
+  --description-list-content-gap: var(--kbq-size-l);
```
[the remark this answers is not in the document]
KamilEmeleev: These are already private. In this repo, private variables are `--<component>-*` and public ones are `--kbq-*`.

packages/components/src/components/DescriptionList/components/DescriptionListGroup/DescriptionListGroup.tsx
KamilEmeleev: Да, нужен. Список строится из групп, как коллекции в React Aria: у группы есть `id`, так работают и статичный список, и список из `items`.

Ещё группа даёт больше свободы в оформлении: пару «термин — описание» можно выделить целиком, например фоном, цветом рамки или подсветкой при наведении. Сетка при этом не ломается: группа встраивается в колонки списка через `subgrid`, и термины всех групп стоят в одной колонке.

`div` внутри `dl` — валидный HTML.

lskramarov (approved):

rmnturov (approved):

PR #486, opened 2026-09-15
fix(ContentPanel): build resizing on Resizable (DS-5422)

KamilEmeleev:
Builds `ContentPanel` resizing on the shared `Resizable` component instead of its own `useMove` logic.

**Fixes**

- A percentage `width` / `defaultWidth` is resolved once the container is measured, instead of falling back to `400px`.
- The panel shrinks together with its container.
- Dragging the handle in RTL grows the panel in the right direction.
- Turning `isResizable` off drops the managed width.
- A double click returns the panel to the width it was opened with, and it keeps following the container afterwards.

**Behavior changes**

- The handle is a `role="separator"` with `aria-valuenow` / `aria-valuemin` / `aria-valuemax`. It still isn't reachable with the keyboard.
- `document.body[data-resizing]` is no longer set during a drag — only the panel gets `data-resizing`. CSS keyed on `body[data-resizing='true']` has to move to the panel.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * ContentPanel resizing now supports pixel and percentage widths, container-aware sizing, minimum/maximum bounds, controlled values, and RTL layouts.
  * Double-clicking the resize handle resets the panel width, with localized accessibility labeling.
  * Consumer-provided styles now take precedence over managed Resizable styles.

* **Bug Fixes**
  * Improved resizing behavior when panels shrink or reach size limits.
  * Resize handles without a maximum no longer announce an artificial maximum value.
  * Resizing state and keyboard interaction now provide more accurate accessibility behavior.

* **Documentation**
  * Added guidance for resizing ContentPanel and resetting its width.


packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
```diff
@@ -1,152 +1,181 @@
...
+    (size: ResizableSize) => onResizeEnd?.(Math.round(size.width)),
+    [onResizeEnd]
+  );
 
-        return next;
-      });
+  return {
+    resizableProps: {
+      // The panel only manages its inline size, the block size is left to CSS.
+      size: isResizable ? { width: panelWidth } : undefined,
```
lskramarov: **[correctness] Turning `isResizable` off leaves a stale inline width**

When `isResizable` turns false, `size` becomes `undefined`. That switches Resizable's `useControlledState` from controlled to uncontrolled, so it falls back to its internal `{ width }` state and keeps setting an inline `width` on a panel that is no longer resizable.

**Failure scenario:** Render `<ContentPanel isResizable defaultWidth={400}>`, drag to 450, then re-render with `isResizable={false}` (e.g. `isResizable={!isMobile}`). In @react-stately/utils 3.11, `stateValue` starts as `{width: 400}` and every `setManagedSize` updates it, even while controlled. So `useResizable` still sets `style.width = 450px`, the `.base { inline-size: 50% }` default never comes back, and dev mode logs "A component changed from controlled to uncontrolled". The old hook returned `width: undefined` here.

```suggestion
      size: isResizable ? { width: panelWidth } : {},
```

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
```diff
@@ -1,152 +1,181 @@
...
+    );
 
-    setWidth(next);
-  }, [isResizable, onResetResize, min, max, setWidth]);
+    const nextWidth = onResetResize?.(initialWidth);
 
-  const { moveProps } = useMove({
-    onMoveStart() {
-      if (!isResizable) return;
+    setUserWidth(
```
lskramarov: **[correctness] Reset pins a clamped px width, so the default stops following the container**

Double-click reset stores the resolved, clamped pixel width in `userWidth`. After a reset, the panel stops following a percentage `defaultWidth` or the container as it did before the first drag, and a reset in a narrow container locks in the clamped width.

**Failure scenario:** `defaultWidth="50%"`, container 1000 → panel 500. Drag to 600, double-click → `setUserWidth(500)`. Container grows to 1600 → panel stays 500 instead of 800. Or `defaultWidth={700}`, reset while the container is 500 → `userWidth=500`; container back to 1000 → panel stays 500 instead of 700. When `onResetResize` returns nothing, reset could go back to the unmanaged state (`userWidth = null`, calling `onResize(initialWidth)` directly) or store the unclamped value.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
```diff
@@ -1,152 +1,181 @@
...
+  const panelWidth = clamp(
+    userWidth ?? resolveWidth(defaultWidth) ?? DEFAULT_WIDTH,
+    min,
+    max
+  );
 
-    const initial = initialResetRef.current ?? 0;
-    const nextRaw = onResetResize?.(initial);
+  // The width the panel started with, resolved again when it's reset.
+  const initialWidthRef = useRef(width ?? defaultWidth);
```
lskramarov, line 157: **[correctness] Live `defaultWidth` and the mount-time reset snapshot disagree**

The displayed width re-reads `defaultWidth` on every render (`userWidth ?? resolveWidth(defaultWidth)`), but the reset target is a mount-time snapshot in `initialWidthRef`, so the two disagree once `defaultWidth` changes.

**Failure scenario:** Mount with `defaultWidth={400}`, then pass `defaultWidth={600}` (e.g. derived from a breakpoint) → the panel renders 600. Double-click the handle → it resets to 400, a width it never showed. After the first drag, later `defaultWidth` changes are silently ignored, so the prop acts as controlled until the user touches it. Either snapshot the default for display too (standard `default*` semantics) or read the current `width ?? defaultWidth` at reset time.

packages/components/src/components/ContentPanel/ContentPanel.tsx
```diff
@@ -123,9 +101,8 @@ const ContentPanelComponent = forwardRef<ContentPanelRef, ContentPanelProps>(
     const rootProps = mergeProps<ComponentPropsWithRef<'div'>[]>(
       {
         ref: panelRef,
-        className: clsx(s.base, isResizable && s.resizable, className),
+        className: clsx(s.base, className),
         style: {
```
lskramarov, line 105: **[correctness] Resizable styles now override consumer `style` sizing**

Resizable merges `style={{ ...styleProp, ...resizableStyle }}`, so the managed `width`/`minWidth`/`maxWidth` now override the consumer's `style`. Before, ContentPanel spread `...style` after `inlineSize`, so the consumer's value won.

**Failure scenario:** `<ContentPanel isResizable style={{ inlineSize: 600 }} />` or `style={{ maxWidth: 'calc(100% - 48px)' }}` used to win. Now the inline `width: 400px` (declared after `inline-size` in the same block) and `max-width: <containerWidth>px` take precedence, silently ignoring the consumer's sizing.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
```diff
@@ -1,152 +1,181 @@
...
-      return {
-        tabIndex: -1,
-        'aria-hidden': 'true',
-      } as HTMLAttributes<HTMLElement>;
-    }
-
-    const aria: HTMLAttributes<HTMLElement> = {
+    handleProps: {
+      direction:
+        direction === 'rtl' ? HANDLE_DIRECTION_RTL : HANDLE_DIRECTION_LTR,
```
lskramarov, line 207: **[correctness] Handle direction comes from the locale, layout from the DOM `dir`**

The handle's physical direction comes from `useLocale().direction`, but the panel is placed with logical CSS (`inset-inline-end`, `margin-inline-start`) that follows the DOM `dir`. When they disagree, the handle lands on the wrong edge and dragging is inverted.

**Failure scenario:** An app with `<Provider locale="ar-AE">` (or no `locale`, so a browser locale like he-IL is used) but no `dir="rtl"` on the document: the layout stays LTR with the panel pinned right, yet `direction=[1,0]` puts the handle at `right: 0`, on the outer edge where the container's `overflow: hidden` clips it, and dragging right grows the panel. The old resizer used `inset-inline-start`, so it always sat on the visible inner edge.

packages/components/src/components/ContentPanel/ContentPanel.tsx
```diff
@@ -166,16 +137,19 @@ const ContentPanelComponent = forwardRef<ContentPanelRef, ContentPanelProps>(
...
+          <Resizable
             {...rootProps}
+            {...resizableProps}
             data-transition={transition}
             data-resizable={isResizable || undefined}
           >
             <Dialog {...dialogProps} ref={overlayRef} role="dialog">
-              {isResizable && <div {...resizerProps} />}
+              {isResizable && (
+                <Resizable.Handle disableKeyboardResize {...resizerProps} />
```
lskramarov, line 148: **[accessibility] Handle announces `aria-valuemax="9007199254740991"`**

The handle now always sets `aria-valuemin`/`aria-valuemax`. A standalone panel with no `maxWidth` announces `aria-valuemax="9007199254740991"`, where the old resizer left these out when `minWidth`/`maxWidth` weren't set.

**Failure scenario:** Standalone `<ContentPanel isResizable>` with no container and no `maxWidth` → `max = Infinity` → Resizable sets `aria-valuemax` to `Number.MAX_SAFE_INTEGER`, which screen readers announce as a meaningless range.

packages/components/src/components/ContentPanel/ContentPanel.tsx
```diff
@@ -145,13 +122,7 @@ const ContentPanelComponent = forwardRef<ContentPanelRef, ContentPanelProps>(
...
     );
 
-    const resizerProps = mergeProps(
-      {
-        className: s.resizer,
-      },
-      slotProps?.resizer,
-      moveProps
-    );
+    const resizerProps = mergeProps(slotProps?.resizer, handleProps);
```
lskramarov: **[correctness] `slotProps.resizer['aria-label']` is always overridden**

`slotProps.resizer` is merged before `handleProps`, so a consumer's `slotProps.resizer['aria-label']` is always replaced by the localized "Resize panel".

**Failure scenario:** `slotProps={{ resizer: { 'aria-label': 'Resize preview' } }}` renders `aria-label="Resize panel"` because `mergeProps` keeps the last non-function value. Reversing the order lets slot props override the label, and handlers still chain.

```suggestion
    const resizerProps = mergeProps(handleProps, slotProps?.resizer);
```

packages/components/src/components/ContentPanel/ContentPanel.module.css
```diff
@@ -22,11 +21,7 @@
   }
 }
 
-.resizable {
-  inline-size: unset;
-}
-
-body[data-resizing='true'] .base {
+.base[data-resizing] {
```
lskramarov, line 24: **[removed-behavior] `body[data-resizing]` is no longer set during a drag**

Dragging no longer sets `document.body.dataset.resizing`. That attribute was visible in the DOM, so app CSS keyed on `body[data-resizing='true']` silently stops matching; now only the panel gets `data-resizing`.

**Failure scenario:** A consumer rule like `body[data-resizing='true'] { cursor: ew-resize; user-select: none }`, used to keep the resize cursor and block text selection while the pointer leaves the 16px handle, no longer matches during a drag. It was undocumented, but it deserves a changelog note.

packages/components/src/components/ContentPanel/__tests__/ContentPanel.test.tsx
```diff
@@ -88,7 +102,7 @@ describe('ContentPanel', () => {
       </ContentPanel>
     );
 
-    expect(getPanel()).toHaveStyle({ inlineSize: '500px' });
+    expect(getPanel()).toHaveStyle({ width: '500px' });
   });
 
   it('should reset width on resizer double click', async () => {
```
lskramarov, line 108: **[test-coverage] Deleted hook tests left the clamp and reset cases uncovered**

Deleting `useContentPanelResize.test.tsx` removed the only tests for clamping the `onResetResize` return value, clamping at min/max during a drag, and "the reset target is the first controlled `width`, not a later one". None of the new component tests cover these paths.

**Failure scenario:** A regression that stops clamping `onResetResize(() => 999)`, lets a drag go past `maxWidth`, or resets a controlled panel to its latest `width` would still pass CI.

packages/components/src/components/ContentPanel/__tests__/ContentPanel.test.tsx
```diff
@@ -109,17 +123,128 @@ describe('ContentPanel', () => {
...
+    expect(resizer).toHaveAttribute('aria-orientation', 'vertical');
+    expect(resizer).toHaveAttribute('aria-label', 'Resize panel');
+    expect(resizer).toHaveAttribute('aria-valuenow', '400');
+    expect(resizer).toHaveAttribute('aria-valuemin', '200');
+    expect(resizer).toHaveAttribute('aria-valuemax', '800');
+    expect(resizer).toHaveAttribute('aria-controls', getPanel().id);
+    expect(resizer).toHaveAttribute('tabindex', '-1');
+  });
+
+  it('should localize the resize handle and flip it in RTL', () => {
```
lskramarov: **[test-quality] The RTL test claims localization but never asserts it**

"should localize the resize handle and flip it in RTL" only checks `data-direction-x/y`. Nothing checks localization, and `intl.json` has no he-IL translation.

**Failure scenario:** The localized `aria-label` could break (e.g. a wrong `ru-RU` key) without failing this test, even though its name says localization is covered.

packages/components/src/components/ContentPanel/components/ContentPanelContainer/ContentPanelContainer.module.css
```diff
@@ -37,7 +37,9 @@
 }
 
 /* panel */
-.panel {
+
+/* Nested in the container to win over the `position` of the Resizable root. */
+.base .panel {
```
lskramarov, line 42: **[altitude] The specificity bump patches Resizable's `position` rule**

Changing `.panel` to `.base .panel` only works around Resizable's unconditional `.base { position: relative }`. Any other code that positions a ContentPanel through `className` (standalone usage) still competes with that rule on stylesheet order.

**Failure scenario:** Standalone `<ContentPanel className={styles.drawer}>` with `.drawer { position: fixed; inset-inline-end: 0 }`, where that CSS loads before the library's `style.css` (or dev injection order differs), gets `position: relative` and renders in flow. Lowering Resizable's specificity (`:where(.base)`) or leaving `position` to the consumer fixes this for everyone and makes the nested selector unnecessary.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
lskramarov, line 207: **[conventions] `refactor` title on a PR that changes behavior**

The PR is titled `refactor(ContentPanel): …`, but it changes behavior: RTL drag direction flips, a percentage `defaultWidth` is resolved after the container is measured, the panel shrinks with the container, the handle gains `role="separator"` and inline min/max widths, and `body[data-resizing]` is gone.

**Failure scenario:** AGENTS.md (included by CLAUDE.md) defines "`refactor` | Code change with no behavior change" and says "Only `feat` and `fix` appear in the changelog". These user-visible fixes will be left out of the generated changelog; the title should be `fix(ContentPanel): …`.

packages/components/src/components/ContentPanel/ContentPanel.tsx
```diff
@@ -166,16 +137,19 @@ const ContentPanelComponent = forwardRef<ContentPanelRef, ContentPanelProps>(
     const panel = (
       <Transition {...transitionProps}>
         {(transition) => (
-          <div
+          <Resizable
```
lskramarov, line 140: **[efficiency] Non-resizable panels still mount Resizable and its observers**

Every ContentPanel now mounts `Resizable`, even when `isResizable` is false and nothing is managed. That still costs a ResizeObserver with rAF state updates, a `useId` id, a context provider, and `position: relative`.

**Failure scenario:** A non-resizable panel in a container that changes size (window resize, body transition) re-renders `Resizable` on every observed size change for nothing. One option is to render the plain root when not resizable; if you do, keep the element type stable so toggling `isResizable` doesn't remount the dialog contents.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
```diff
@@ -1,152 +1,181 @@
...
+const DEFAULT_MIN_WIDTH = 200;
+
+/**
+ * The handle sits at the inline start edge of the panel, and `Resizable`
+ * directions are physical, so the horizontal axis flips in RTL.
+ */
+const HANDLE_DIRECTION_LTR: ResizableHandleDirection = [-1, 0];
+const HANDLE_DIRECTION_RTL: ResizableHandleDirection = [1, 0];
+
+const clamp = (value: number, min: number, max: number) =>
```
lskramarov: **[reuse] The local `clamp` duplicates existing helpers**

The hook declares its own `clamp`, but `@koobiq/react-core` already re-exports `clamp` from `@react-stately/utils`, and `Resizable/hooks/utils.ts` has another copy.

**Failure scenario:** That makes three copies of the same helper, and they already differ: react-stately's `clamp` is `Math.min(Math.max(v, min), max)` and returns `max` when `min > max`, while this one returns `min`. Choose one behavior on purpose instead of letting the copies drift.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
```diff
@@ -1,152 +1,181 @@
...
+  );
 
-    const initial = initialResetRef.current ?? 0;
-    const nextRaw = onResetResize?.(initial);
+  // The width the panel started with, resolved again when it's reset.
+  const initialWidthRef = useRef(width ?? defaultWidth);
 
-    const next = clamp(isNumber(nextRaw) ? nextRaw : initial, min, max);
+  const onDoubleClick = useCallback(() => {
+    const initialWidth = clamp(
```
lskramarov: **[simplification] `onDoubleClick` duplicates the width resolve-and-clamp logic**

`onDoubleClick` repeats the same resolve → `DEFAULT_WIDTH` fallback → clamp steps as `panelWidth`, and calls `parseContentPanelSize` directly instead of using the `resolveWidth` helper written for this.

**Failure scenario:** The width-resolution rule now lives in two places that must stay in sync (fallback default, min/max clamp). One `resolvePanelWidth(value)` helper used by both would remove the duplication.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
KamilEmeleev: Fixed. `size` is now an empty object when the panel isn't resizable, so `Resizable` stays controlled and drops the inline width instead of keeping the last one.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
KamilEmeleev: Fixed. When `onResetResize` returns nothing, the panel goes back to the unmanaged width and `onResize` is called with the initial width, so it keeps following the container and a percentage `defaultWidth` as before the first resize.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
KamilEmeleev: Fixed. The mount-time snapshot is now used for display too, so later `defaultWidth` changes are ignored and reset returns the width the panel was showing. The snapshot keeps the raw prop, so a percentage still resolves once the container is measured.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
KamilEmeleev: Renamed the PR to `fix(ContentPanel): build resizing on Resizable (DS-5422)`, thanks — the user-visible fixes belong in the changelog.

packages/components/src/components/ContentPanel/ContentPanel.tsx
KamilEmeleev: Fixed, slot props are merged last now, so `slotProps.resizer['aria-label']` wins and handlers still chain.

packages/components/src/components/ContentPanel/__tests__/ContentPanel.test.tsx
KamilEmeleev: Added tests for all three: the clamped `onResetResize` return value, the min/max bounds during a drag, and resetting a controlled panel to the width it was mounted with.

packages/components/src/components/ContentPanel/ContentPanel.tsx
KamilEmeleev: Fixed in `Resizable`: the managed style is merged first, so an inline `style` from the consumer wins again. Covered by a test there and one in `ContentPanel`.

packages/components/src/components/ContentPanel/ContentPanel.tsx
KamilEmeleev: Fixed in `Resizable`: `aria-valuemax` is only set when the size has a real maximum, so an unbounded handle no longer announces `9007199254740991`. `aria-valuemin` stays, since 0 is the true lower bound.

packages/components/src/components/ContentPanel/components/ContentPanelContainer/ContentPanelContainer.module.css
KamilEmeleev: Keeping the nested selector. Lowering `position: relative` to `:where(.base)` would let any consumer class win it by accident, and the absolutely positioned handles depend on it. The nested selector only reaches the panel the container places.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
KamilEmeleev: Fixed, the hook uses `clamp` from `@koobiq/react-core` now. `max` is kept at or above `min`, so the bounds can't cross and both implementations agree.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
KamilEmeleev: Fixed, both paths go through a single `resolvePanelWidth` helper.

packages/components/src/components/ContentPanel/__tests__/ContentPanel.test.tsx
KamilEmeleev: Split in two: one test asserts the flipped direction in RTL, the other the `ru-RU` label.

packages/components/src/components/ContentPanel/ContentPanel.module.css
KamilEmeleev: Added a note to the PR description.

packages/components/src/components/ContentPanel/hooks/useContentPanelResize/useContentPanelResize.ts
KamilEmeleev: Leaving it as is. `useLocale().direction` is what `Tooltip` and `Popover` use in this library, and React Aria expects the app to keep `dir` and the locale in step.

packages/components/src/components/ContentPanel/ContentPanel.tsx
KamilEmeleev: Split out as DS-5586. `isDisabled` wouldn't help today — the observer runs either way — so `Resizable` will skip measuring while resizing is disabled, which helps every consumer. Rendering a plain root instead would remount the dialog whenever `isResizable` flips.

